PR #23, opened 2015-09-25
magnetodynamics: fix joule field

juhanikataja:


juhanikataja:
More specifically, fix joule field calculation in 2D

PR #25, opened 2015-10-14
Parallel tests

samiilvonen:
Initial support for parallel tests with one example test case.


juhanikataja:
The `TEST.PASSED` file generated by each test have the same name regardless of number of PEs which causes IO race condition when ctest is run the `-j` option.

PR #31, opened 2015-11-03
Last merge from devel to release before 8.1

juhanikataja:


juharu:
```
* Moi 
* jos on B/H käyrä, niin valittaa CalcFields:stä (jossa ilmeisesti kysellään tuota joka tapauksessa...) 
* 
* 
* Warn if permeability or permittivity is not given in magnetodynamics solver 
```

PR #33, opened 2015-11-05
ElmerGrid: Added PF3 file import

pavelponomarev:
First version of PF3 import from Cedrat Flux 3D


pavelponomarev:
It is not yet tested and there might be problems with dimension and order of elements. Will debug further next week.

pavelponomarev:
The code is tested with elements 303, 306, 404, 408, 706 and 715.

PR #38, opened 2015-11-17
Wedges mesh multiplication

pavelponomarev:
Enables splitting of first order prism(wedge) elements
Tested in serial case


juharu:
Pavel, could you please remove this segment from the code (was just for debugging)

+if ( facenumber > SIZE(Eparent  % FaceIndexes) ) THEN
+stop 'notound'
+endif

PR #40, opened 2015-11-25
Namespace

juharu:
Is this a good & sound idea? Anyway, has found its use already....


juharu:
Breaks the BlockLinElast3 -test, where block -iterative solvers now get their setting
from "outer: " -namespace, not from the default namespace, as the test assumes...

PR #51, opened 2016-03-29
Multiple workarounds to enable compilation with older versions of Int…

mbycklin:
…el Fortran.


mbycklin:
The code should now compile with ifort 15 when AVX2 compilation is being performed.

PR #64, opened 2016-09-01
Cmake openmp

juhanikataja:
- OpenMP related MPI stubs were added enable compilation of Elmer with OpenMP but _without_ MPI
- OpenMP cmake compiler test was made stricter so that it fails if `!$OMP SIMD` directive is not understood
- `PoissonThreaded2` and `PoissonThreaded3` tests were disabled if OpenMP is enabled because they break the compilation of Elmer


mbycklin:
OpenMP SIMD construct should always be followed by a canonical loop, see http://www.openmp.org/mp-documents/openmp-4.5.pdf, Section 2.8.1. The test could be of the form
REAL :: A(16), B(16)
!$OMP SIMD
DO i=1,16
 A(i)=B(i)
END DO
!$END OMP SIMD

juhanikataja:
Fix in [#65](https://github.com/ElmerCSC/elmerfem/pull/65)

PR #65, opened 2016-09-05
FIX: canonical loop after OMP SIMD construct in FindOpenMP.cmake

juhanikataja:


mbycklin:
Looks better. I just noticed that my original end loop construct was the wrong way around (not required by the standard so goes unnoticed by the compiler): it should be, if included !$OMP END SIMD

PR #72, opened 2016-11-02
Masked structure fix

drschrank:


raback:
Eliminated as agreed with drschrank.

PR #75, opened 2016-11-18
Harmonic 2D torque fix

pavelponomarev:
Fixed 2D harmonic torque computation with help of Anouar and Jenni


juharu:
ThanKs Pavel!

juhanikataja:
I modified the nodal force for harmonic case so that it gives similar torque as with the airgap torque formula developed in [47f5b68](https://github.com/ElmerCSC/elmerfem/pull/75/commits/47f5b68fc62a3c1e59b76bea503206ff05c3b36d). A test case in `fem/tests/mgdyn2D_em_harmonic` but it only results in a torque that is in a same ballpark with the airgap formula. The commit is [here](https://github.com/ElmerCSC/elmerfem/commit/15109271edec4ba4e4a2178c02d31c1376c68c23).

PR #80, opened 2016-12-20
Threading and SIMD improvements

mbycklin:
Threading improvements:
- Function GetCurrentElement should now return the current element pointer
when called from within OpenMP parallel regions
- Changed keyword Multithreaded Solver to MultiColour Solver (Logical)
- Added keyword MultiColour Consistent (Logical), which enables consistent
colorings to be created (same coloring irrespective of the number of threads
used).
- Fixed and enabled PoissonThreaded, PoissonThreaded2 and PoissonThreaded3
tests

Performance improvements:
- Cache blocked ElementInfoVec for better efficiency when the number of
points to be computed is large
- Cache blocked most routines in LinearForms module
- Removed copyin/copyout from calls to basis functions in ElementInfoVec
- Added GetReadValues subroutine for implementing the same functionality as
GetReal, but without copying the values to a temporary location as an
intermediate step

New test: LinearFormsAssembly for testing the local matrix assembly of
linear forms for all H1 elements except pyramids.


raback:
Nice developments that will make it much easier to create multithreded solvers!

PR #84, opened 2017-01-18
Elmergui qt5

juhanikataja:


juhanikataja:
This will allow ElmerGUI to be compiled with qt5 using WITH_QT5=TRUE cmake flag. If the flag is not given, revert back to qt4.

PR #93, opened 2017-03-08
Fix for axis-symmetric time harmonic problems

jvencels:


ettaka (approved):
There is a question about the sign convention. In cartesian coordinates the current density is towards \hat{e}_z and in cylindrical coordinates \hat{e}_theta these are towards opposite directions.

PR #94, opened 2017-03-13
Allow explicit coil axes for CoilSolver

juhanikataja:
Controlled with keywords

* `Coil Normal (3,1) = real nx ny nz`
* `Coil Tangent (3,1) = real tx ty tz`


juhanikataja:
Now it suffices to give only the coil normal: `coil normal(3,1) = real nx ny nz`

PR #100, opened 2017-05-31
SIMD and Threading enhancements

mbycklin:
Further SIMD and threading improvements.

**SIMD**
- SIMD improvements to GetElementNOFDOFs.
- Improved SIMD performance of ElementInfoVec, ElementMetricVec and LinearForms for a small number of Gauss points.
- Significantly improved the performance of CRS_GlueLocalMatrixVec by swithing to an alternative algorithm and introducing software prefetching (software prefetching currently requires Intel Fortran compiler).
- H1Basis has been refactored to avoid register spilling arising from accessing unaligned multidimensional array.

**Threading**
- Completely removed locking from critical path in FE assembly, improving multithreading efficiency.
- Improved NUMA locality by initializing the system matrix in parallel with threads.
- Improved NUMA locality by making Gauss points structure private to each thread.
- ElementInfoVec and LinearForms module now use stack storage for work space.
- GetCurrentElement and SetCurrentElement modified to give out correct values when called from within parallel regions.
- Modifications to enable experimental multithreaded startup. New functionality can be used with the Solver keyword _Multithreaded Startup = Logical <True|False>_
- Modified CheckElementEquation to be thread safe.

**General improvements**
- Partially refactored ElementInfoVec to enable support for other element types than p elements in the future.
- Simplified fetching of local dofs for edges and face when element has a uniform polynomial degree.
- Corrected SParIterSolver to call CRS_MatrixVectorMultiply (for real matrices).
- Minor code fixes (uninitialized variables).
- Added testing for different values of element polynomial degree p to basis function tests.
- Modifications to BiCGStabL to avoid temporary copies when calling LAPACK.
- Modifications to ShellSolver to avoid temporary copies on calls to internal subroutines.
- Reverted DefaultUpdateEquationsR interface back to accept noncontiguous arguments.
- Minor corrections to Lists.F90 to correct uninitialized variables
- Added recursive attributes to routines where needed.


juhanikataja:
Was there an issue with this one. Can this be pulled?

mbycklin:
ctest -L quick passes without errors, but some of the longer tests do fail. I requested a review by @juharu. I wonder if there is an update on that?

PR #107, opened 2017-08-02
Forced steady state using EEC

pavelponomarev:


raback:
If EEC is False does not affect existing code - safe to merge.

ettaka:
Please consider the following:
1)	Do you think the CircuitsAndDynamics module is the correct place for this vector potential variable manipulation?
I didn’t have the access to this paper (can you please send it?) but what I read from the code is that 
you correct the magnetic vector potential by the negative average of the initial and half period solution.
I think this should be done in the MagnetoDynamics2D (and/or) Magnetodynamics modules right after the solution
has been reached (possibly after the nonlinear loop). Could you please consider this? I think putting it in the 
circuitsoutput is a bit hackish and can be very misleading and may even cause some trouble later on.
2)	this is just a style issue. But I believe when Info is called, one should put the subroutine that calls it as the first string. For example now you have  Info('CircuitsAndDynamicsEEC', "Performing EEC at half-period.", Level=1) but there is no such a routine called ‘CircuitsAndDynamicsEEC’.

PR #116, opened 2017-10-06
cmake: dont install metis, just link it statically in ElmerGrid

juhanikataja:


samiilvonen:
Only thing that worries me is that at some point someone turns on dynamic linking for metis and then we have conflict with system version. Should we also force the static linking?

PR #119, opened 2017-10-16
Collection of multithreading improvements.

mbycklin:
Collection of multithreading improvements. Also some minor fixes regarding code correctness.

Multithreading improvements:

- Added multithreading to HUTI double precision CG algorithm.
- Added partial implementation of **ListMatrixArray** type and module. **ListMatrixArray** which offers the same operations as **ListMatrix** but allows thread safe access per matrix row. **ListMatrixArray** also pools **ListEntry** allocations per thread, which is more time efficient with multiple threads than allocating **ListEntries** one at a time (less locking in the Fortran runtime and OS).
- Modified CreateMatrix to use **ListMatrixArray** and multicoloring when constructing the initial matrix sparsity structure and multithreaded startup is enabled.
- Added partial multithreading to selected matrix scaling routines.
- Added multithreading to matrix bandwidth computation. Bandwidth optimization routines not multithreaded due to serial nature of the used RCM algorithm.

Bug fixes:

- Corrected race conditions arising from implicit SAVE variables from MagnetoDynamicsUtils and MagnetoDynamics.
- Corrected OpenFOAM IO module to work with the Intel Fortran compiler.
- Corrected SaveLine output level to use the maximum of 31
- Corrected uninitialized variable access in ElmerGrid

Misc. improvements:

- Boundary mesh is now also colored in AddEquationBasics when Keyword MultiColour Solver is set to True. Colour index lists for boundary mesh are available in Mesh % BoundaryColourIndexList similarly to regular colour index lists.
- Modified MakePermUsingMask to use List_AllocateMatrix

### DIFF
--- a/elmergrid/src/femelmer.c
+++ b/elmergrid/src/femelmer.c
@@ -5116,7 +5116,8 @@ int SaveElmerInputPartitioned(struct FemType *data,struct BoundaryType *bound,
   int *bcnodesaved[MAXBCS],maxbcnodesaved,*bcelemsaved,*bcnode;
   int *bcnodedummy,*elementhalo,*neededtimes2;
   int partstart,partfin,filesetsize,nofile,nofile2,nobcnodes;
-  int halobulkelems,halobcs,savethis,fail,cdstat;
+  int halobulkelems,halobcs,savethis,fail=0,cdstat;
+
   FILE *out,*outfiles[MAXPARTITIONS+1];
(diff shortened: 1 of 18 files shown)


raback:
Great developments!

juhanikataja:
Compiles and tests pass nicely on cluster as well with intel compilers 17.0.4.

PR #126, opened 2017-11-30
New linear forms and multithreading enhancements.

mbycklin:
Multithreading improvements:

- Added multithreaded and mesh colored version of ModelPDE. The case has been
  slightly modified from original ModelPDE to better meet the Elmer coding
  standards.

- Improved multithreaded performance by removing unnecessary initializations
  of some allocatable strings.

- Added multithreading to HUTI BiCGStab, HUTI GMRES and BiCGStabL (double
  precision only).

- Added multithreading to HUTI dummy preconditioning function (double
  precision).

- Added multithreading to dot product and two norm computations with MPI

- Added multithreading to selected loops in SparMatrixVector in SParIterSolver
  module.

- Improved NUMA locality of the solution vector by initializing the vector in
  parallel.

SIMD improvements:

- Added linear forms for (grad u, v) and (u,v) in H^1 to LinearForms module.

- Added testing of (grad u, v) and (u,v) linear forms to LinearFormsAssembly
  test (only constant coefficients).

Bug fixes:

- Corrected arguments in call to DGEMM for simultaneously projecting multiple
  vectors to FE basis vectors.

- Set minimum and maximum output PE to be within PE bounds in a parallel
  enviroment.

- Corrections to uninitialized or out of bounds accesses to variables
  (StressSolver, MainUtils)

- Corrected GetElementNodesVec to set ElementNodes pointers properly and zero
  out memory which is not used.

- Corrected ElementInfoVec to give correct result in the case given basis
  vector set could contain more entries than needed.

Misc. improvements:

- Revised LinearFormsAssembly test to loop through several polynomial
  degrees. Also modified the test to use the correct amount of integration
  points per element / polynomial degree.

- Added List_AddMatrixIndex -function for adding elements to list matrices
  with a 0 initial value (does not affect the value of existing
  elements). This function is now called from various places within the
  solver to avoid unnecessary computations.

- Removed call to MKL dcsrgemv from CRS_ABSMatrixVectorMultiply to avoid stack
  overflows for large matrices when the code is used with MKL.


raback:
Again, great new developments from mbyclin! All tests pass that passed earlier.

PR #128, opened 2018-01-05
A collection of minor patches to correct some test issues.

mbycklin:
- Fixed multiple out-of-bound accesses.
- Fixed potential divisions by zero via masking.
- Fixed some undefined pointer references.
- Fixed some uninitialized variable accesses.
- Increased linear solver accuracy to make some tests pass. Actual values to
  test against unchanged
- Changed direct solver to UMFPACK in selected tests as MUMPS might
  not be available by default.


mbycklin:
I think the changes proposed to **Lists.F90** and **SolverUtils.F90** would need to be reviewed before merging. Some of the tests do still fail, but the results seem rather close to tolerances in most of the cases.

PR #129, opened 2018-01-10
mistype the == in for(;;) in Line 3665

gwjwin:


raback:
Thanx for the fix! I committed the fix just to devel but will not merge this one as there seems to be some unneeded changes as well.

PR #135, opened 2018-02-20
Build system enhancements

mbycklin:
- Improved the way CMake detects and uses BLAS/LAPACK from Intel Math
  Kernel library.

  BLAS and LAPACK routines from Intel MKL are now used by default if
  detected by FindMKL unless BLAS_LIBRARIES and LAPACK_LIBRARIES CMake
  variables have been set. This is to avoid the use of CMake
  FindBLAS/FindLAPACK -routines to find MKL, as those tend to find
  incorrect MKL core libraries for GNU Fortran, resulting in runtime
  failures.

- Minor corrections to library link order in FindMKL.
- Minor corrections to internal structure of FindMKL, which should now work
  better with the QUIET option (Windows build still untested)

- Added detection of OpenMP SIMD features within the build
  system. Added routines for checking the existence and the
  functionality of the used OMP SIMD -features in the code.

  Added global preprocessor macros to allow OpenMP SIMD functionality
  to be disabled if needed:

  When OpenMP >=4.0 is available, the following mappings are valid:
  - _ELMER_OMP => $OMP
  - _ELMER_OMP_SIMD => $OMP SIMD
  - _ELMER_OMP_DECLARE_SIMD => $OMP DECLARE SIMD

  When OpenMP >=4.5 is available, the following mappings are valid:
  - _ELMER_LINEAR_REF(var) => LINEAR(REF(var))


raback:
Great! More robust build system is appreciated by many.

PR #137, opened 2018-04-11
Change string list "cvalue"  from "character" to "varying_string".

raback:


raback:
This is such an old pull request that the initial reason for it is inclear. So maybe just forget about it...

PR #146, opened 2018-06-14
Zirka devel inject

juhanikataja:
Zirka-Moroz hysteresis model for MagnetoDynamics 2D incl. 3 tests:
* circuits2D_with_hysteresis: 2d test with circuit
* circuits2D_with_hysteresis_axi: 2d axisymmetric test with circuit
* Zirka: unit test that tries to recover hysteritic BH curve from FE simulation.


juhanikataja:
Doing merge from devel first.

juharu (approved):
OK!

PR #151, opened 2018-06-28
cmake: make OpenMP 4.5 test stricter

juhanikataja:
The following causes gfortran (7.3 on ubuntu 18.04 and possibly 8 too) to die in `(null):0: confused by earlier errors, bailing out`:

```fortran
module test_module
contains
  subroutine minimal(u)
    real :: u
    !$OMP DECLARE SIMD LINEAR(REF(u))
  end subroutine
end module

program test_prog
use test_module

print *, "test prog"
end program
```
Compiled with `gfortran-7 -fopenmp -v -save-temps test.f90 -o test`

Same thing will happen with ElementDescription.F90 and H1Basis.F90.


mbycklin:
After commit 16501400aafb328c230c0e61631ed658da73dc08, the test for OpenMP 4.5 fails always due to USE statement of TESTOPENMP45MODULE being incorrectly positioned within the TESTOPENMP45 program unit. 

Could you try moving the USE statement before the IMPLICIT NONE and trying again if the test catches the compiler issue you are after?

mbycklin:
OpenMP 4.5 test now works as expected with the Intel Fortran compiler (tested both with 2018u2, 2018u3). 

However, with ifort the compilation itself fails in multiple places within ZirkaHysteresis.F90, to what I believe are actually code issues. 

>fem/src/ZirkaHysteresis.F90(520): error #8300: If a dummy argument is allocatable or a pointer, and the dummy or its associated actual argument is polymorphic, both dummy and actual must be polymorphic with the same declared type or both must be unlimited polymorphic.   [RC_P]
  rc_p => RecurseDepth(rc_p, B)
-----------------------^

To solve this, one could change `rc_p` to be defined as `TYPE(RevCurve_t), POINTER :: rc_p`. The pointer assignment should work as `rc_p` and `rp` are type compatible. 

Another compilation failure is encountered on lines 729 and 736.

> fem/src/ZirkaHysteresis.F90(729): error #5521: A derived-type object in an input/output list cannot have inaccessible components unless a suitable user-defined input/output procedure is available.
    if (.not. present(rc0)) print *, X, rc_p % depth, c_loc(rc_p), c_loc(rc_p % parent)
----------------------------^

`C_LOC` intrinsic returns `TYPE(C_PTR)` which cannot be output as such. To make the code compile, one could, for instance, convert C_PTR to an integer via `TRANSFER(c_loc(rc_p),adum)`, where `adum` is defined as 

`integer(kind=addrint) :: adum`

However, this is technically not guaranteed to output the address of `rc_p`, although with most compilers it probably will.

juhanikataja:
Thank you for commenting and testing. The issues were indeed real coding issues, now (hopefully) fixed in PR #153.

PR #154, opened 2018-07-02
Lua code inside sif files

juhanikataja:
* Includes Lua 5.1.5 interpreter inside Elmer codebase under
  contrib/lua-5.1.5
* Enabled in compilation with cmake variable `WITH_LUA`.
    * Setting `USE_SYSTEM_LUA=TRUE` makes cmake look for system Lua.
    * CMake variables `LUA_LIBRARIES`, `LUA_INCLUDE_DIR` disables cmake
      from searching Lua.
* Enables Lua expressions inside sif file in following cases
    * Inline syntax with `#<lua expression>#` or `#<lua expression>`
      similarly to matc
    * Commented sections in main sif file as follows
      ```
      !---LUA BEGIN
      ! <first line of lua code>
      ! <second line of lua code>
      ! ...
      ! <last line of lua code>
      !---LUA END
      ```
      Such code blocks are to be executed prior to reading the rest of
      the sif-file. Thus, such code blocks are not executed in included
      files.
    * Using variable dependent keyword evaluations:
      ```
      keyword = variable var_a, var_b, var_c, ..., var_n
        real lua "expr"
      tensor keyword (i,j) = variable var_a, var_b, var_c, ..., var_n
        real lua "expr 1, expr 2, expr 3, ..., expr i*j"
      ```
      Here the entries to the `tensor keyword` are given in a row-wise
      order.
* Should work in threaded mode too.
* Includes 2 tests: `Lua` and `KeywordUnitTestLua`


fem/src/Lua.F90
```diff
@@ -0,0 +1,386 @@
...
+  character(kind=c_char) :: fname(*)
+  real(kind=c_double), intent(in) :: X(:)
+  real(kind=c_double), intent(inout) :: Y(:)
+
+  integer :: nx, ny, i, lstat
+  nx = size(X,1)
+  ny = size(Y,1)
+  CALL lua_getfield(L%L, LUA_GLOBALSINDEX, fname)
+  do i = 1,nx
+    CALL lua_pushnumber(L%L, X(i))
```
mbycklin, line 320: With ifort, the compilation fails here, in the `lua_pushnumber` interface first argument is 
`type(c_funptr), value :: L` whereas `L % L` is `type(c_ptr)`.

mbycklin (changes requested):
Does not compile with Intel Fortran 2018u2, fails at fem/src/Lua.F90:309 due to argument type on call to lua_pushnumber not matching.

mbycklin (approved):

fem/src/Lua.F90
```diff
@@ -0,0 +1,386 @@
...
+  if(c_associated(L%L)) L % initialized = .true.
+end function
+
+subroutine lua_addfun(L, fun, fname)
+  type(LuaState_t) :: L
+  procedure(luafun), pointer :: fun
+  type(c_funptr) :: c_fun
+  character(kind=c_char):: fname
+
+  c_fun = transfer(fun, c_fun)
```
mbycklin: I don't see the point in using `transfer` in this case. In addition, such a line seems to crash ifort. It would be better to replace the line with `c_fun = c_funloc(fun)` as it achieves the same functionality via standard Fortran methods.

mbycklin (changes requested):
Suspicious use of transfer in fem/src/Lua.F90 causes ICE with ifort.

juhanikataja:
Using `transfer` also crashed ifort `17.0.4 20170411` and changing to `c_funloc` corrected the problem.

mbycklin (approved):

PR #163, opened 2018-11-06
Crsvec fix

juhanikataja:
There was possibly an indexing bug in `CRS_GlueGlobalMatrixVec` when ndofs > 1 and masking was not needed.

I changed the "masking needed" branch to follow same indexing for consistency.

Please disregard CMakeLists.txt stuff. That's from another commit.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,6 +43,8 @@ SET(WITH_CONTRIB FALSE CACHE BOOL "Include contributed solvers")
 
 SET(WITH_LUA FALSE CACHE BOOL "Include LUA extensions.")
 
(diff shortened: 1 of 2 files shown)


mbycklin:
The indexing in multi DOF case with no masking seems incorrect like you pointed out. 

Instead of the original
`(NDOFs*N)*(i-1)+NDOFs*(j-1)+cdof`
the linear index to `Lind` and `Lvals` should be
`(N*NDOFS)*(NDOFs*(i-1)+rdof-1)+NDOFs*(j-1)+cdof`
but I believe that the proposed fix of manually maintaining the count should work equally well.

mbycklin (approved):
Looks functional to me. To test against regressions, a simple test case for gluing process with vectorization and ndofs>1 might be worth to implement.

PR #167, opened 2018-12-05
Calving-GlaDS-Plume Changes

Morlocke:
The majority of the changes required to set up a combined calving, hydrology and plume simulation. Should be up-to-date with the current code base.


raback (changes requested):
Hi, It seems that you have completely new versions of HeatSolve, MeshSolve, FlowSolve under elmerice. It seems redundant. I guess the changes to these solvers are so small that they could remain in the generic directory (/fem/src/modules)?

Morlocke:
Hi Peter.

Yeah, the change is just to get the solvers to pick up the new MeshTag
field when deciding whether they need to reallocate their memory at the
very start. Because of the timestepping I'm using, Mesh % Changed isn't
reliable, so I've had to modify all the solvers I need to pick up the new
field. They should all still pick up Mesh % Changed as well, so it should
all be backwards compatible.

Thanks

Samuel

raback:
Hi Samuel, 

Yes, I understand. However, there are completely new redundant solvers copied from fem/src/modules to elmerice/Solvers directory. See for example

https://github.com/ElmerCSC/elmerfem/pull/167/files#diff-3a4b4800abcb6323fdd58a250d73377c

These should, in my opinion, remain under fem/src/modules only.

best regards, 
Peter

Morlocke:
Hi Peter

Oh, I see - sorry, I misunderstood your first email. Yes, I think I just
wasn't paying attention when I moved things into the upload directories.
I'll sort that out.

Thanks

Samuel

raback:
I made some changes to the devel branch such that SingleSolver library routine will check the MeshTag and compare it to Solver % MeshTag. This means that there is no need to solver-specific checks any more. Hence when making the merge we should mainly be able to use the devel version of non-ice related solvers.

Morlocke:
OK, great. That should simplify things on the merge, then, as it means a
lot of the modified solvers can be ignored.

raback:
Merged in another way.

PR #171, opened 2018-12-17
Perform MPI tests in CI

juhanikataja:
Also include

* Zoltan
* Mumps

Refactor `opts*.cmake` to `opts-generic.cmake` under `ci/precaches/` for platform agnostic options.


juhanikataja:
Tests pass after merge so I'm merging this.

PR #177, opened 2019-03-27
Misc. typo and whitespace fixes

luzpaz:


ElmerGUI/Application/plugins/egmain.cpp
```diff
@@ -2,7 +2,7 @@
    ElmerGrid - A simple mesh generation and manipulation utility  
    Copyright (C) 1995- , CSC - IT Center for Science Ltd.   
 
-   Author: Peter R�back
+   Author: Peter R�back
```
luzpaz, line 5: sorry about this. Using a graphical git client and there must be a bug since it's not showing up in my UI. Will try to revert these changes

luzpaz: darn, I can't revert these changes since it doesn't show up in my UI :angry:  
Not sure how to do this via the CLI

raback:
Thanx!

PR #183, opened 2019-08-04
ElmerGUI bugs

t7saeki:
Hi,

This is a proposal to fix some bugs in ElmerGUI.

**1. A bug described in  [a topic in Elmer Discussion Forum](http://www.elmerfem.org/forum/viewtopic.php?f=15&t=2608&p=8725&hilit=improving#p8028)** 
- _removal of boundary conditions or materials (if not last one in list?) leads sometimes to unstable state: boundary conditions / materials cannot be removed any more, boundary condition box will not close. User has to save the project and restart ElmerGUI before continuing to work._

When removing a material (or boundary condition, equation, initial condition, body force - all same), the DynamicEditor instance is removed from MainWindow::materialEditor.  This causes inconsistency between DynamicEditor::ID of other DynamicEditor instances in materialEditor and their position in materialEditor while DynamicEditor::ID is used as the position in materialEditor  like below. (the second argument id in below example is the the ID)

> // signal (int,int) emitted by material editor when ready:
> //-----------------------------------------------------------------------------
> void MainWindow::matEditorFinishedSlot(int signal, int id)
> {
>   DynamicEditor *pe = materialEditor[id];
>   

So, instead of removing from matrialEditor, I put negative number (-100) to DynamicEditor::ID as DynamicEditor instances which have negative ID are always skipped though the program.  However, this change causes a jump in indexes in egproject.xml which crushes ElmerGUI when loading the egproject.xml. So, I added a code to adjust the indexing in MainWindow::saveProjectContents(...) function.

**2. Another bug described in the same topic**

- _after removal of a body force, the removed body force is still written in the sif file. Workaround: save project, restart ElmerGUI._

This happens when generating sif file before BodyPropertyEditor::ui.bodyForceCombo is updated by calling MainWindow::populateBodyComboBoxes(BodyPropertyEditor*). I manually let the combo select 0th item just after the removal in MainWindow::bodyForceEditorFinishedSlot(int, int).  (same for material, equation, initial condition, boundary condition) 

**3. Boundary conditions mixed up when edges/surfaces are unified after attaching boundary conditions to edges/surfaces**
This is caused by redundant (or unused) BoundaryPropertyEditor instances remaining in MainWindow::boundaryPropertyEditor and solved by just removing the redundants in MainWindow::edgeUnifySlot() and MainWindow::surfaceUnifySlot().

Hope this helps improve ElmerGUI. 

Best regards.
Saeki

### DIFF
--- a/ElmerGUI/Application/src/mainwindow.cpp
+++ b/ElmerGUI/Application/src/mainwindow.cpp
@@ -1784,6 +1784,7 @@ void MainWindow::saveProjectContents(QDomDocument projectDoc,
 
   QDomElement editorBlock = projectDoc.createElement(blockName);
   projectDoc.documentElement().appendChild(editorBlock);
+  int index = 0; // index excluding removed DynamicEditor instances
 
   for(int i = 0; i < Nmax; i++) {
     DynamicEditor *de = editor[i];
@@ -1793,7 +1794,7 @@ void MainWindow::saveProjectContents(QDomDocument projectDoc,
 
     // Menu item number:
     QDomElement item = projectDoc.createElement("item");
-    item.setAttribute("index", QString::number(i));
+    item.setAttribute("index", QString::number(index++));
     editorBlock.appendChild(item);
 
     // Is active?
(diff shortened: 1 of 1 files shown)


raback:
Great work! This is really appreciated as there has not been much resources to keep up ElmerGUI which is still quite usable educational tool. Thanx!

PR #192, opened 2019-11-30
GUI project loading

t7saeki:
Hi all,

This is a proposal of ElmerGUI improvement mainly in project loading 

**Changes around project loading**
- Added recent project list in the file menu for quick loading (up to 5 projects)
- Fixed bug on doubled "Solver specific options" tab by reloading which was discussed in http://www.elmerfem.org/forum/viewtopic.php?f=4&t=2113
- Auto solver loading from /edf-exra directory when loading a project. This relieves users from manually appending extra solvers prior to loading the project. 
- Added menu item to suppress the auto sif generation (Sif->Suppress auto generation). This allows users to protect manually modified sif contents. If turned on, the existing sif file will be loaded when loading project, and text contents in the sif editor will be saved to .sif file when saving mesh and sif-files/project.  

**Other changes**
- Added project folder path information in the title bar
- Storing/restoring window position of main window, sif editor, solver log and convergence graph
- Hidden sharp edges by default in Windows environment to avoid crush by double clicking boundaries
- Adjusted Setup dialog size to make Apply button visible even in low resolution display (may not enough but at least much better for my display...)

Hope you like it.  

Best regards.
Saeki


raback:
Hi Seaki, 

Great effort! This is really appreciated!

BR, Peter

PR #193, opened 2019-12-28
GUI syntax highlighting

t7saeki:
Hi there,

This is a proposal of GUI improvement:
- Solver log window class was separated from Sif window class to implements below changes. Previously, both are instances of same SifWindo class.
- "Preference" menu was added to sif window and solver log window for font selection and syntax highlighting options. This is to make GUI fun to work with.
- "Generate, save and run" button was added to main window (Green doubled triangle button). This button generates sif file and saves project then runs solver. This is useful to quickly run a new condition modified via GUI.
- "Save and run" button was added to sif window (Green triangle button). This button saves the sif window contents as .sif file,  saves project (without generating new sif) and runs solver. This is useful to quickly run manually edited sif code.
- Window geometry restoration is simplified by using QWidget::saveGeomery() / QWidget::restoreGeomery(). (The previous code was very complicated.)  

Hope you like this.

Regards,
Saeki


raback:
t7saeki, nice work again! Could you contact me at raback (at) csc.fi

PR #194, opened 2020-01-26
Gui object browser

t7saeki:
Hi all,

This is another proposal for GUI improvement:

- Object Browser was added on the left side of main window.  The Object Browser is a tree view that contains information on geometry(bodies and boundaries) and model(equations, materials, body forces, initial conditions and boundary conditions), and it provides users easy access to these. This is for a request in Elmer Discussion Forum (2nd bullet point of "New Features" in http://www.elmerfem.org/forum/viewtopic.php?f=15&t=2608&p=8725#p8028) 

	- Single click of each body/boundary item in the Object Browser highlights corresponding object in the mesh, and double click pops up body/boundary property editor for the body/boundary.
	- Single click of each equation/material/etc. item in the Object Browser highlights corresponding object in the mesh where the equation/material/etc. was set, and double click pops up the editor for the equation/material/etc..
	- Double click of "[Add...]" next to Equation/Material/Etc. group pops up a new editor for equation/material/etc. 

  Users can hide/show the Object Browser from the menu "View -> Show Object Browser". This is for a workaround when the Object Browser causes sever problem.

- Body/boundary checkboxes in equation/material/etc. editor are synchronized to mesh highlighting. (This works only when the Object Browser is shown.)

- Buttons for "Menu -> Set body properties" and "Menu -> Set boundary properties" were added to main window. (next to "Unify edge" button)

- A scrollbar was added in Setup dialog to make sure that Apply button is accessible.

Hope you like it.

Regards,
Saeki


raback:
Wau Saeki! That's a huge improvement. Great work! Compiled for me out of the box, and also old projects loaded. Will merge with pleasure.

PR #198, opened 2020-03-09
elmergrid: tentative fix for comsol 55 mphtxt reader

juhanikataja:


raback:
Point to vertex. Thanx Juhani!

PR #215, opened 2020-04-11
GUI: Removal of auto sif generation and "Use as a body" checkbox

t7saeki:
Hi,

This is a proposal for some modification in ElmerGUI

- Removal of auto sif generation during loading/saving project
 Conventional ElmerGUI automatically generate sif when loading/saving project and overwrite existing sif which was annoying. In the past I added a menu to suppress the auto sif generation, but now I want to remove the auto sif generation completely for simplicity.  Instead, users can generate sif file anytime by "Generate" command. or can use "Generate, save and run" command for convenience. 
This is just my preference. If you see any benefit of the auto sif generation, please let me know. 

- Removal of "Use as a body" checkbox  
"Use as a body" shouldn't be for a single boundary but for a boundary condition which can include multiple boundaries. In the future, this checkbox may be moved to DynamicEditor for Boundary Condition, but for now the checbox is just hidden and disabled to avoid misleading users.  

Hope you like it.

Regards,
Saeki

### DIFF
--- a/ElmerGUI/Application/src/boundarypropertyeditor.cpp
+++ b/ElmerGUI/Application/src/boundarypropertyeditor.cpp
@@ -48,6 +48,8 @@ BoundaryPropertyEditor::BoundaryPropertyEditor(QWidget *parent)
   : QDialog(parent)
 {
   ui.setupUi(this);
+  ui.boundaryAsABody->setEnabled(false);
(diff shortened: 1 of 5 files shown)


raback:
Indeed, it seems that "use as body" is in wrong place.

PR #216, opened 2020-04-12
Wip/metis

Foadsf:
Now the METIS library is installed via MSYS2/MinGW-w64 packages


Foadsf:
@raback I think the Travis-ci YAML and the docker machines need to be updated.

PR #218, opened 2020-04-15
devel2calving

raback:
We should merge devel to calving before vice versa.


raback:
Merging with different strategy.

PR #265, opened 2021-03-23
MPI_init and _finalize only if requested + Bugfix in i2s

Azrael3000:
2 small changes that are as follows:

1. MPI_Init only if required

   When using Elmer as a library from a software that already runs MPI calling MPI_Init again is not working, so to enable this check if MPI is already initialized and skip Init if it is already there. In this case we also don't want to call MPI_Finalize as the other  oftware should take care of that.

2. Shorten i2s

    The old version would give wrong results if the numbers became too large due to a division by zero.

If you have any questions, let me know.

### DIFF
--- a/fem/src/SParIterComm.F90
+++ b/fem/src/SParIterComm.F90
@@ -132,6 +132,7 @@ CONTAINS
     ParEnv % MyPE = 0
     ParEnv % PEs  = 1
     ParEnv % ActiveComm = 0
+    ParEnv % ExternalInit = .FALSE.
 
     ierr = 0
 #ifdef _OPENMP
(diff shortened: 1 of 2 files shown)


juharu:
Hi Arno!
Thanks for the patches. The MPI -part is fine with me. As for the I2S part, I seem to remember that some compilers have, at least historically, had trouble dealing with i/o statements within i/o statements. That might be the reason why I2S() is what it is.
This change would at least need testing with different compilers?  Maybe you could push the MPI part separately ?
Br, Juha

juharu:
Hi Arno,
I pushed a fix for I2S to "devel". Maybe you could test it out ?
-Juha

Azrael3000:
I rebased to your new devel and cherry picked the MPI commit only.

I think your i2s fix should work now. It's a bit hard to reproduce (I saw it while doing a valgrind run, that's why the numbers were so high).

juharu:
Merged. 
Thanks for the patch!

PR #269, opened 2021-04-03
GUI fix - crash by checkbox

t7saeki:
Hi all,

This change is to solve crashing issue of ElmerGUI on Windows, reported in[ Elmer Discussion Forum ](http://www.elmerfem.org/forum/viewtopic.php?p=23724#p23724).  In more detail, this issue is caused by an additional itemClicked signal after itemDoubeClicked signal.  Interestingly, I didn't see such crash in Ubuntu, but only in Window.

To avoid the crashing, [Add..] button on Object browser was changed to work by a single click instead of doubleclick. Hope this helps.

Regards,
Saeki


raback:
Great work Saeki!

PR #272, opened 2021-04-14
WIP - Add more GUI menu XMLs

kdunn926:
I'm planning to add a load of XMLs for missing modules in this PR. I was hoping someone could do a once-over on this first commit and make sure I'm on the right track.

### DIFF
--- a/ElmerGUI/Application/edf-extra/advection-reaction.xml
+++ b/ElmerGUI/Application/edf-extra/advection-reaction.xml
@@ -1,166 +1,166 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE edf>
 <edf version="1.0">
-	<PDE Name="Advection-Reaction">
-		<Name>Advection-Reaction</Name>
-		<Equation>
-			<Parameter Widget="Label">
-				<Name> Options </Name>
-			</Parameter>
-			<Parameter Widget="Combo">
-				<Name>Convection</Name>
-				<Type> String </Type>
-			</Parameter>
-			<Parameter Widget="TextEdit" Enabled="True">
-				<Name> Free text </Name>
-				<Type> String </Type>
-				<Whatis> Free text is copied into the Equation-block of the SIF as such. </Whatis>
-				<StatusTip> Free text is copied into the Equation-block of the SIF as such. </StatusTip>
-			</Parameter>
-			<Item>
-				<Name>None</Name>
-			</Item>
-			<Item>
-				<Name>Computed</Name>
-			</Item>
-			<Item>
-				<Name>Constant</Name>
-			</Item>
-		</Equation>
-		<Solver>
-			<Parameter Widget="Label">
-				<Name> Advection-Reaction Solver </Name>
-			</Parameter>
-			<Parameter Widget="Edit" Enabled="True">
-				<Name> Procedure </Name>
-				<DefaultValue> "AdvectionReaction" "AdvectionReactionSolver" </DefaultValue>
-			</Parameter>
-			<Parameter Widget="Label">
-				<Name> Options </Name>
-			</Parameter>
-			<Parameter Widget="Edit">
-				<Name> Equation </Name>
-				<Type> String </Type>
-				<DefaultValue> Advection Reaction Equation Tracer </DefaultValue>
-				<Whatis> The name of the equation. </Whatis>
-				<StatusTip></StatusTip>
-			</Parameter>
-			<Parameter Widget="CheckBox" Enabled="True">
-				<Name> Discontinuous Galerkin </Name>
-				<Type> Logical </Type>
-				<Whatis></Whatis>
-				<StatusTip> Needs to be set to true. </StatusTip>
-			</Parameter>
-			<Parameter Widget="Edit">
-				<Name> Variable </Name>
-				<Type> String </Type>
-				<DefaultValue> Tracer </DefaultValue>
-				<Whatis> The name of the variable. As the variable is a DG variable (i.e., not renderable e.g. in ElmerPost), the user usually adds the option -nooutput in order to avoid output in the output files. </Whatis>
-				<StatusTip></StatusTip>
-			</Parameter>
-			<Parameter Widget="Edit">
-				<Name> Steady State Convergence Tolerance </Name>
-				<Type> Real </Type>
-				<Whatis> With this keyword a equation specific steady state or coupled system convergence tolerance is given. All the active equation solvers must meet their own tolerances for their variable c before the whole system is deemed converged. The tolerance criterion is: ||ci − ci−1|| &lt; e||ci ||, where e is the value given with this keyword. </Whatis>
-				<StatusTip></StatusTip>
-			</Parameter>
-			<Parameter Widget="CheckBox">
-				<Name> Limit Solution </Name>
-				<Type> Logical </Type>
-				<Whatis> Assumes the variational inequality method to apply, if set to true. </Whatis>
-				<StatusTip></StatusTip>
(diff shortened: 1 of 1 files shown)


raback:
Hi, 

It would be highly appreciated if you would write some of the missing XML's! 

I suggest that you study the XML file for HeatSolver. It includes a lot of the same complexity that is needed here. See, https://github.com/ElmerCSC/elmerfem/blob/devel/ElmerGUI/Application/edf/heatequation.xml

For example, the "convection velocity" and "exported variable" are missing the index. It should be "convection velocity 1" etc. There is fancy "activate" stuff also available. This means that some entries activate new fields. This is often just luxury but some keywords use it. For example, in HeatSolver the "Exported Variable 2" is activated only after "Exported Variable 1" is filled. 

The "Nonlinear System" and "Linear System" keywords are common to all solvers. 

When I've written some of these, my logic is as follows:
1) Take some a little similar module as starting point
2) Take the chapter in Models Manual and rename/remove/add the keywords
3) Make some simple test

Hope this helps!

richb2k:
If I may make a suggestion, it's probably a good idea to add 'free text input' sections in Equations, Solvers, and Boundary Conditions at a minimum.  This way if there is a needed keyword or value that is not listed in the EDF, then the user can easily add the keyword.  It's possible to add 'free text input' to every section, but these three sections are probably the most useful.

kdunn926:
> I suggest that you study the XML file for HeatSolver. It includes a lot of the same complexity that is needed here. See, https://github.com/ElmerCSC/elmerfem/blob/devel/ElmerGUI/Application/edf/heatequation.xml
> For example, the "convection velocity" and "exported variable" are missing the index. It should be "convection velocity 1" etc. There is fancy "activate" stuff also available. This means that some entries activate new fields. This is often just luxury but some keywords use it. For example, in HeatSolver the "Exported Variable 2" is activated only after "Exported Variable 1" is filled.

Thanks for the pointer @raback .

> The "Nonlinear System" and "Linear System" keywords are common to all solvers.

My takeaway from this is any keyword starting with [Linear System, Nonlinear System] can be omitted from the XML, right?

> When I've written some of these, my logic is as follows:
> 
> 1. Take some a little similar module as starting point
> 2. Take the chapter in Models Manual and rename/remove/add the keywords
> 3. Make some simple test
> 
> Hope this helps!

This certainly helps. I've only used Elmer for slightly beyond hello world at this point and am not familiar with any of the models really, so I would be surprised if I could identify any similarities aside from a name. I was hoping to be able to more or less blindly work through the Models Manual and drop in keywords in the right sections, etc. 

Given my inexperience with Elmer, for a "simple test", does something as basic as trying to open an XML qualify or should the XML be used to drive through a hello world simulation?

kdunn926:
> If I may make a suggestion, it's probably a good idea to add 'free text input' sections in Equations, Solvers, and Boundary Conditions at a minimum. This way if there is a needed keyword or value that is not listed in the EDF, then the user can easily add the keyword. It's possible to add 'free text input' to every section, but these three sections are probably the most useful.

I very much welcome any and all suggestions @richb2k . I have incorporated a 'free text input' for all sections.

richb2k:
As Peter mentioned, making a simple test of the new .xml file is key to making a working EDF menu file.  Here's one approach:  Since the goal is to develop advection-reaction.xml that will generate a working sif file, one can work backwards from a working sif that was manually generated and compare the output from elmersolver for both sif files.  

For example, there are several working advection-reaction test cases that work in 'elmerfem/fem/tests', try to use your new xml file to create a similarly working sif file.  

I tried this using your xml file with the test case 'AdvReactDB', but without MPI, and found a few issues, such as:
CheckKeyword:  Unlisted keyword: [variable_name gamma] in section: [material 1]
CheckKeyword:  Unlisted keyword: [variable_name source] in section: [body force 1]
CheckKeyword:  Unlisted keyword: [variable_name] in section: [boundary condition 1]
where it should not say 'variable_name', but should say something like 'Potential'.  See attached test case.
[test-xml-adv-react.zip](https://github.com/ElmerCSC/elmerfem/files/6321343/test-xml-adv-react.zip)

I've found that it will probably take 3 to 4 attempts, before you get a working xml file.

Rich.

kdunn926:
Thank you Rich. This testing suggestion makes perfect sense (and in hindsight should have been obvious).

Regarding the specific discrepancies for advection-reaction, I do see all the tests specify `Potential`, rather than the `variable_name` placeholder. The only remaining question I have is to understand how to know `Potential` should be have been used instead of `variable_name`, based on the advection-reaction section in the Models Manual (i.e. the term "Potential" is never used in that section)? Should I read this simply as an oversight in the manual and always refer back to test cases when it looks like specifics for keywords are omitted?

richb2k:
All I can say for sure is that variable_name doesn't work, elmersolver doesn't like it.  Looking at other xml files, such as Electrostatics, might help.  IMHO the Models Manual should be the last place to doubt.  

At this point, it might be better to start a conversation in the ElmerGUI forum about this topic.

I agree with Peter, it's great you want to work on adding more EDF files for the other solvers.  I was thinking about adding a chapter to the ElmerGUI manual describing what you are currently doing, so this is a perfect way to start writing up a procedural outline.

kdunn926:
It looks like I have a working sif file generated using the latest committed EDF menu file.

My plan was to work through the models manual in order and create menu files for each missing model. Does this seem reasonable for a single PR or should I split the work up more?

richb2k:
I ran a test with the latest xml file, and it looks like there are too many free text fields.  For example, in Materials, the entry for 'Convection Velocity 1' acts like a free text field, instead of assigning a value to a variable.  One is supposed to only enter 'Variable Coordinate   Real MATC "ux(tx)" ' and not 'Convection Velocity 1 = Variable Coordinate   Real MATC "ux(tx)" ' into the box.  The user is not supposed to enter the variable names.  Compare the generated case.sif to the original sif file.
[test-xml-adv-react.zip](https://github.com/ElmerCSC/elmerfem/files/6330108/test-xml-adv-react.zip)

So with a little editing, the new xml file should be ready for action.
Rich.

raback:
Why the potential was initially "varname" is because it can be arbitrary. This then affects the Dirichlet conditions etc. so we need to fix it for the GUI. Perhaps "Potential" is not the best one. It is already reserved for some solvers. Even "Tracer" is better. "Concentration" is a little long, and not always accurate either. Maybe "Substance". When we choose it, we have to live with it...

richb2k:
Searching through the Models Manual for the word 'Tracer' returned 3 results, and all three are in the Advection-Reaction chapter.  Searching through 'Solver.Keywords', returned zero results.  Any place else to search for 'Tracer'?

raback:
If we choose "Tracer" that should be added in SOLVER.KEYWORDS so that the code does not complain. That goes for other keywords of ElmerGUI as well. The user should not need to fill the type of the keyword in ElmerGUI.

raback:
I took this XML file, modified it somewhat + modified the code quite a bit to make the graphical use slicker. For instance, the nodal field is now obtained by ticking a box rather than introducing an exported variable. The default name is also now "Tracer". Also keywords were added to SOLVER.KEYWORDS. I only dry tested it (i.e. looked that the .sif file looks good to the eye) but didn't try to solve any cases. The AdvReact basic test case was also updated to be consistent. Old cases should also work fine.

raback:
So unfortunately I will not merge this after the manual work. Thanx for everybody who worked on this! The updated stuff is on branch "AdvReactRevisit".

richb2k:
Great, I'll give it a try tomorrow using the nightly build.  And thanks to kdunn926 for starting the work!

PR #283, opened 2021-06-28
GUI geometry handling by mouse

t7saeki:
This PR is on geometry handling by mouse drag based on [request](http://www.elmerfem.org/forum/viewtopic.php?p=8028#p8028) in Elmer Discussion Forum.

- Zoom by right mouse button
- Pan by left+right mouse button. This is for Windows/Linux laptops without mouse

The menu by right mouse button click is still available, but the menu appears only when right mouse button is released where it was pressed (i.e. without moving mouse.)


raback:
Great new feature Saeki!

PR #284, opened 2021-07-03
Custom material libraries

t7saeki:
This PR is for the request in [Elmer Discussion Forum](http://www.elmerfem.org/forum/viewtopic.php?p=6238#p6238).
Added options for custom material libraries by locating xml files (written in same manner as egmaterials.xml) in edf-extra directory. User can select the xml file in the comboBox which was newly added at the top of material library dialog.


raback:
Nice feature!

PR #288, opened 2021-09-20
fix build with Qt5Gui

sprhawk:
Failed to build ElmerGUI on Debian Bullseye, it reported
```
error: aggregate ‘QPainterPath path’ has incomplete type and cannot be defined
```
It needs to `#include <QPainterPath>` to pass build


raback:
This fix has been committed directly among other developments. Hence closing the ticket without merging. Thanx for reporting!

Issue #276, opened 2021-05-08
ElmerGUI does not handle  Output, Timestep intervals and Timestep Sizes arrays


ealbiter:
Some weeks ago, I tried to use variable time steps in a transient simulation, following the suggestions in this [forum post](http://www.elmerfem.org/forum/viewtopic.php?p=5097&sid=f29a652fa6cfec6684a9129fd40c14f9#p5097). However, I realized that ElmerGUI does not take into account the n values entered in these fields since the generated SIF does not contain the correct size of the array, i. e., it does not generate the Output Intervals(n), Timestep Intervals(n) and Timestep Sizes(n) lines correctly. Due to this, ElmerSolver only uses the first value of these fields.
I have located the lines that need to be modified in the SIFgenerator class and [added my proposed changes](https://github.com/ealbiter/elmerfem/commit/29104315b961d12f1189f29ccfe2b18fa9a335f9) to correct this issue. If you wish, I can make a PR for your review. 
By the way, thank you for this excellent software.

raback:
Yes, this would make life easier! Thanx for your effort. Please make a PR. 

Currently you can use the following syntax:
`Timestep Sizes = Size 3; 0.01 0.1 1.0`
The problem is that this is unintuitive and nobody uses this.

PR #291, opened 2021-10-05
Fix issue #276

ealbiter:
As described in #276, this PR allows to handle arrays in the Output, Timestep intervals and Timestep Sizes variables.

### DIFF
--- a/ElmerGUI/Application/src/sifgenerator.cpp
+++ b/ElmerGUI/Application/src/sifgenerator.cpp
@@ -177,15 +177,18 @@ void SifGenerator::makeSimulationBlock()
 	     ui.simulationTypeCombo->currentText().trimmed());
   addSifLine("  Steady State Max Iterations = ",
 	     ui.steadyStateMaxIterEdit->text().trimmed());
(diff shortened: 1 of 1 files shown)


raback (commented):
Hi 

Thank you for your efforts! I was about to accept this but I'm a little worried that somebody uses accidentally more than one white spaces. I must admit that this is probably rare and would probably be fine. I'm not a Qt hack myself but how would you alternatively see using a temporal variable: 

QStringList pieces = qs.split( " " );

Then the number of pieces should be available in "pieces.length()". The splitting goes fine even though there would be accidental white spaces.

ealbiter:
Hi Peter, and thank you for your feedback. Indeed, this code does not handle well the additional white spaces.
I will try your approach to handle this issue, but, at the moment, my concern is that the use of the split() function would increase the QT requirement to v5.14, according to the [online documentation](https://doc.qt.io/qt-5/qstring.html#split).
By the way, I also have very little experience in QT programming, but I will try to propose a better solution.

raback:
Well done!

Issue #293, opened 2021-10-15
Fail to build with Qwt 6.2


ealbiter:
Qwt was recently updated to v6.2 in my system. With this version, ElmerGUI fails to compile due to the qw_compat.h file was [removed](https://qwt.sourceforge.io/qwtchangelog.html).
I have already made the changes to resolve the problem. I tested the fix with Qwt 6.1.6 and 6.2.0, and the code compiles fine with both versions. In a short time, I will make a PR for your review.

raback:
Thanx! Just merged.

ealbiter:
Happy to help. Closing issue.

PR #294, opened 2021-10-15
Compatibility with Qwt 6.2. Fix for #293

ealbiter:
This modification should be compatible with Qwt 6.X. I don't know about older versions, but I think that shouldn't be a problem because v6 has been available in several distros since several years ago (including LTS ones).


raback:
Seems legit. I tested with two platforms and faced no issues. Thank you for contributing!

PR #297, opened 2021-11-03
GUI segmentation fault by doubleclicking

t7saeki:
I noticed that windows version of ElmerGUI crashes by doubleclicking mesh when compass (or numbers, etc) is shown. This PR is to avoid such crash by just hiding them when handling double click.


raback:
Great fix Saeki!

PR #301, opened 2021-11-12
Qt5 detection

t7saeki:
Just added Qt5 detection in CMakeLists.txt to skip manual definition of WITH_QT5. I'm not so sure this is helpful for all the people, but is convenient at least for myself who uses cmake-gui.


raback:
Thanx Saeki for your continued efforts! This will make compilation easier for many.

PR #303, opened 2021-11-13
Fix local Find*.cmake overloading the CMake ones.

### LINKED ISSUE #299: Conflicts with `cmake:3.20`.
What is the rationale of keeping local copies of CMake internal modules i.e:
- FindPackageHandleStandardArgs.cmake
- CMakeParseArguments.cmake
- FindMPI.cmake
They break build setup in later CMake versions.

bartoszek:
Fix #299

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,7 +88,7 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
 endif(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
 
 
-SET(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/Modules" ${CMAKE_ROOT}/Modules)
+SET(CMAKE_MODULE_PATH "${CMAKE_ROOT}/Modules;${PROJECT_SOURCE_DIR}/cmake/Modules")
 
 SET(CMAKE_Fortran_MODULE_DIRECTORY
   ${PROJECT_BINARY_DIR}/fmodules CACHE PATH "Directory for Fortran modules")


raback:
I had to revert this since I got reports that it caused the tests not running succesfully. I didn't dig deeper being no cmake wizard myself.

bartoszek:
@raback thanks for poking me about this, I'll give it a looksy :wink:

bartoszek:
> I had to revert this since I got reports that it caused the tests not running succesfully. I didn't dig deeper being no cmake wizard myself.

I've pinpointed the issue here, you can't use `${CMAKE_MODULE_PATH}` in `add_test()` directly as it gets expanded and effectively only the first element of the list will get passed to the `COMMAND` arg list.

Check out this simple example: 
```cmake
cmake_minimum_required(VERSION 3.2)
project(test)

set(list test1 test2)
string(REPLACE ";" "\\;" list_escaped "${list}")


message("list: ${list}")
execute_process(COMMAND echo pass list as argument to command: ${list})

message("list_escaped: ${list_escaped}")
execute_process(COMMAND echo pass list_escaped as argument to command: ${list_escaped})
```
You'll get:
```bash
$cmake -S . -B build
list: test1;test2
pass list as argument to command: test1 test2
list_escaped: test1\;test2
pass list_escaped as argument to command: test1;test2
-- Configuring done
-- Generating done
-- Build files have been written to: /tmp/test/build
```

To fix this you can use relative path to `test_macros.cmake` in `fem/tests/*/runtest.cmake` same as in `elmerice/tests/*/runtest.cmake`

https://github.com/ElmerCSC/elmerfem/blob/5fc4b393cdaa99f851a2fa00adf82650c53553d0/fem/tests/1dtests/runtest.cmake#L1-L3

https://github.com/ElmerCSC/elmerfem/blob/5fc4b393cdaa99f851a2fa00adf82650c53553d0/elmerice/Tests/Calving3D/runTest.cmake#L1-L5

Or process `CMAKE_MODULE_PATH` with `string(REPLACE )` to escape semicolons.

PR #308, opened 2021-11-29
[vtk:9.1] fix cmake config and version headers

bartoszek:
Please review vtk components list, I'm not sure if all of those are needed (tested against vtk:8.2.0 and vtk:9.1.0)
And VS Code for Web on Github fucking slaps :smirk: 
image


raback:
This cmake stuff is not really my cup of tea. I tested this on some platforms and happily merged only to find out that on an older one this would not compile. Hence I had to take this back. Trying to cut some corners didn't really pay of here. Sorry for the extra trouble!

bartoszek:
No problemo, I'm here for it.

PR #314, opened 2022-01-22
OCC options

t7saeki:
Enabling to leverage OCCT (PR#311) by bartus is nice. I took the commit from PR#311 and modified so that cmake finds OCE if OpenCASCADE is not found for those who still use OCE. I tested in Ubuntu20.04LTS and MSYS2 both for OCE and OCCT.
Hope this encourages Elmer team to merge PR#311.


raback:
Thank you saeki and  bartoszek!

PR #315, opened 2022-01-23
Some fixed bugs in GUI

t7saeki:
This PR is for fixing some bugs:

**Very small Sif Window and Solver Log Window** 
 Initial window size for Sif Window and Solver Log Window was not set and it caused very small window when launching ElmerGUI for the first time.  This is fixed in the PR by adding initial size for these windows.

**Default directory name for mesh saving** 
Default directory name for mesh saving was not retained when users canceled saving mesh.  This is improved in the PR by storing default directory name in MainWindow::saveAsSlot().
 
**3D mesh display**
In some cases, 3D mesh was not displayed correctly when loading project (typically, GenericTemperature sample in GUI tutorial). This is improved in the PR by adding redraw operation in the end of  ObjectBrowser::loadProjectSlot().

### DIFF
--- a/ElmerGUI/Application/src/mainwindow.cpp
+++ b/ElmerGUI/Application/src/mainwindow.cpp
@@ -8116,3 +8116,11 @@ void MainWindow::selectParaViewSlot(){
   selectVtkPostAct->setChecked(false);
   selectParaViewAct->setChecked(true);
 }
+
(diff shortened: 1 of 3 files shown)


raback:
Great that you keep the fixes coming. Really appreciated by the community!

PR #316, opened 2022-02-23
GUI body selection

t7saeki:
This PR is to mitigate the body selection issue of ElmerGUI discussed in [Elmer Discussion Forum](http://www.elmerfem.org/forum/viewtopic.php?t=7692) . When users double-click a surface shared by multiple bodies, ElmerGUI cannot determine right body to select ( and actually selects nothing.) This causes a problem especially when a body is fully surrounded by another body.  Unfortunately, selecting body by using ObjectBrowser also didn't work in such case. I have added an procedure to 'guess' right body and it also helps select the body by using Object Browser even in such case.
 
This is just a guess based on the assumption that users must hide at least one surface of the outer body to double click the inner body. (This is not always correct as clipping makes it possible to see the inner body without hiding the outer body, but it must be rare case.) The guess may not right in all the cases, but users can select the body by Object Browser.

### DIFF
--- a/ElmerGUI/Application/src/glwidget.cpp
+++ b/ElmerGUI/Application/src/glwidget.cpp
@@ -1041,11 +1041,16 @@ void GLWidget::mouseDoubleClickEvent(QMouseEvent *event)
 	  count++;
 	  found = i;
 	}
+//	cout << i << ": tmp1,2= " << tmp1[i] << "," << tmp2[i] << endl; 
       }
       
       if((count == 1) && (found >= 0))
 	currentlySelectedBody = found;
-      
+    else if((count > 1)){
+	  int m = mostVisibleBody(MAX_BULK_INDEX, tmp1);
+	  if(m >=0) currentlySelectedBody = m;
+	}
+     
       delete [] tmp1;
       delete [] tmp2;
     }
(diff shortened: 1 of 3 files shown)


raback:
Saeki, you're a treasure! I know many people have been struggling with this.

richb2k:
Hello Saeki,  
Regarding your comment 'but users can select the body by Object Browser', using ElmerGUI prior to this PR and using the four body example, clicking on bodies 2 or 4  in the Objectbrowser tree selected properly.  Clicking on bodies 1 or 3 in the Objectbrowser tree causes ElmerGUI to crash because the call to selectBody in Line 462 in objectbrowser.cpp returns a NULL.  selectBody returns a body based on boundaries.  My question is, why call selectBody to look at boundaries, when the user has just double clicked on a body in the tree, such as body 1?  The double click on a body in the Objectbrowser tree is unambiguous, unlike double clicking on the graphic window display of the bodies.  Hope this makes sense, and looking forward to trying out your PR after tonight's build.  Thanks,  Rich.

t7saeki:
Hi Rich,

Thank you for your comment. Your feedback is highly appreciated as testing GUI is a hard work.  Now I see my code lacks NULL checks elsewhere in objectbrowser.cpp. I will add NULL checks to avoid future potential crash and issue PR once ready.

Why call selectBody to look at boundaries? ElmerGUI originally didn't have functionality to highlight a 3D body but highlight only one surface.  To highlight a 3D body, all the surface which belongs to the specified body must be identified - this is a little complicated code because boundaries are not explicitly associated to body in the internal data format of ElmerGUI.  Hope this answers to your question.


Regards,
Saeki

richb2k:
Hello Saeki,

The nightly build dated 2022-02-24 works very well when selecting bodies.  Thanks for the fix!

Best regards,  Rich.

PR #318, opened 2022-03-07
Fix for harmonic losses

jvencels:
Currently "harmonic loss linear" and "harmonic loss quadratic" are showing only losses from imaginary part.


juharu:
Merged. Thanks Juris!

PR #320, opened 2022-03-27
Excess core loss coefficient and interpolation

jvencels:
For electrical steel, core loss depends on frequency and magnetic field flux.
Three coefficient models are more accurate and used by commercial software.
1) hysteresis core loss coefficient,
2) eddy-current core loss coefficient
3) excess core loss coefficient

Additionally, an interpolation scheme is added for the missing element type.


jvencels:
There is sufficient amount of coefficients that can be used for calculation of core losses.
I think there is no need to add another one.

PR #323, opened 2022-04-06
add desktop and metainfo file

kevinsmia1939:
This add desktop file which might be good for ElmerGUI which as use the desktop icon.
The metainfo.xml provide the basic information of the application that could be useful for software store to display relevant info to the user.
https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html

The id is fi.csc.Elmer which is derive from a reverse DNS.
https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#file-naming

This is made when I am packaging Elmer as Flatpak.


raback:
Thanx for contributing!

PR #332, opened 2022-06-28
Add build instructions for Windows

swesonga:
#200 mentions some scripts for building on Windows but they do not appear to be in the Git repository. These instructions explicitly list dependencies that need to be copied into the install directory, but it would be ideal if the script for this operation lived with the source code.


raback:
Thanx Saint! Probably not the final word what comes to compiling but clearly a needed improvement for Windows users.

PR #335, opened 2022-07-23
Fix typos by changing gsmh -> gmsh

swesonga:


raback (commented):
Thanx! Seems ok. Funny that this same typo is appearing so many times.

PR #336, opened 2022-07-24
GUI display scaling

t7saeki:
This PR is to improve some issues seen when displaying GUI with scaled display settings in high resolution monitors.

- Unfixed item height in DynamicEditors . The height of items on QTabWidgets on DynamicEditors had been fixed to 18. This causes problem when seeing with scaled display settings.
scaling_comparison

- Some adjustments of html style in About dialog (replacing ＜BR＞ with ＜P＞＜/P＞) so that all the texts are visible even by scaled display settings

Hope these are useful.

Regards,
Saeki

### DIFF
--- a/ElmerGUI/Application/forms/newproject.ui
+++ b/ElmerGUI/Application/forms/newproject.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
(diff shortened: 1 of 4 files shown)


raback:
Thanx Saeki for your continued support of ElmerGUI!

PR #343, opened 2022-08-06
Fix source code typos

luzpaz:
These commits are various source typo fixes. Please review closely.


raback (commented):
Seems legit. It is more dangerous to fix typos in code but this is limited in scope and most are still comments. Thanx!

PR #347, opened 2022-08-11
GUI updated icons

t7saeki:
Hi,

This PR is to update Oxygen icons used in ElmerGUI to show larger icons when using scaled monitors as shown in the attached picture. The Oxygen icons are separated from the executable binary file of ElmerGUI and installed into share/ElmerGUI/icons/oxygen directory so that ElmerGUI loads Oxygen icons with appropriate size via QIcon::fromTheme() function. 

icon_size_comparison

Regards,
Saeki


raback:
ElmerGUI keeps getting better and nicer looking!

Issue #348, opened 2022-08-16
Save permeability from H-B curve in MagnetoDynamics


ettaka:
SaveMaterials solver does a great job in saving materials as fields.

It seems that the permeability cannot be saved as field in case an H-B curve is used.

To solve this issue, try to implement a way to save the permeability either in SaveMaterials, or directly in MagnetoDynamics that already has the machinery to compute it.

ettaka:
I created a branch [348-save-permeability-from-h-b-curve-in-magnetodynamics-2](https://github.com/ElmerCSC/elmerfem/tree/348-save-permeability-from-h-b-curve-in-magnetodynamics-2) for solving this issue (Still testing if it works well enough)

raback:
Seems legit and the most natural place to add this.

raback:
You might consider having keyword "calculate permeability" instead of "save permeability". We know that it does not really most of the time need to be calculated, or is anyways calculated so in that respect "save" is closer to what happens in the code. However, there are already ~8 fields controlled by "calculate" so having one field with "save" would break the pattern...

ettaka:
@raback you are absolutely right! Thanks for the replys! I will do the modification. Also, we noticed some indexing issue. I will take care of it and we will test it thouroughly before commiting. Thanks a lot!

ettaka:
I added now the modifications as discussed. There is now a test for harmonic case [c2630d9](https://github.com/ElmerCSC/elmerfem/commit/c2630d94f67c1ee64f02479ce7461bd9399a0de2) (however, we will not keep it or then we need to fix the target tolerances)

jvencels:
@raback Could this feature be included in the dev branch?

ettaka:
@raback @jvencels Just one more thing needs to be done: I modified two test cases for demonstration, but I think we need to revert them back. At least one of them and then make sure they do not produce error (reference tolerances change as we add this computation)

raback:
Sure, when this is ready make a PR.

ettaka:
@jvencels FYI https://github.com/ElmerCSC/elmerfem/pull/349

ettaka:
Thanks @raback @jvencels

PR #349, opened 2022-08-18
Implement Calculate Relative Permeability as field

ettaka:
* Calculate Relative Permeability keyword activates the feature
* Computation as 1/Nu/mu0 (tensorial nu not taken into account)
* modified:   fem/tests/mgdyn_bh/case.sif
  * we add a keyword to test the Calculate Relative Permeability feature
  * The test reference norm is changed as the calculated new field is
    entering in the process.

#348


raback:
Nice feature!

PR #355, opened 2022-10-23
MagnetoDynamics, PlatiseFerroModel

platise:
Hi,

According to email conversation with Peter Råback, who provided "function" interface to the magnetic materials I've added my model for use in MagnetoDynamics, WhitneyAVSolver, which can i.e. describe the B-H characteristics of CERN example with 2 dipoles (4 values), and guarantees the function is smooth and monotonic.
Test case is also provided, with a core on a wire, and flux measurement inside the boundary conditions, based on ticket: 
https://www.elmerfem.org/forum/viewtopic.php?t=7820

The "H-B Curve Derivative" keyword did not work, also in my case, the inverse is too complicated to have a direct solution; I've implemented a general solution with n-terms (however 2 of them were in my cases sufficient) based on newton iteration; therefore it would be good if the same function could output derivative besides the H, to avoid the 2nd re-calculation.

BR Uros


raback:
Nice example! However, I would not like merge this as it is. You are mixing the case with the elmerice stuff that is only optionally compiled. We could perhaps have user functions also similarly with the main Elmer, or it could be added within fem/src/modules/MagnetoDynamics/Utils.F90. Also it would perhaps be nice to have the full features case rather here: https://github.com/ElmerCSC/elmer-elmag. This could have pictures etc. also provided. Additionally, we could have just a minimal consistency test run with Elmer. Now this is not automatically run (missing some files for the automation) and anyway requires gmsh. If it is ok, I can make some regrouping.

platise:
I think it should be within the main Elmer; the provided material function could be used in any relation where B-H characteristics is needed. Yes please just do re-grouping, maybe accept pull and then git mv ... and I shall merge back.

Regarding the gmsh, the .msh output from gmsh can be added -- actually I've deleted it afterwards as I thought it is just "spamming" the git.

richb2k:
Running the test case gives this error: ERROR:: LoadInputFile:  Unknown specifier: [2.63089794e2 1.96053253e4], plus two more similar error lines.  Adding the keyword 'Real' twice to coreWires.sif, lines 120 and 121, like this: 'PFM Dipoles Field Strength(2) = Real ....', fixes the error.
Edit: running Windows 10, nightly binary build dated 2022-10-05

platise:
Interesting, on my side no errors --- running latest/devel deb nighty build.

platise:
Another thing that I am trying to understand is the the accuracy/behav of flux density reported to results/coreWire.dat.*

With paraview it can be seen the expected saturated core (2.2 T) is within the core, while surface is at mT, and hence average reported flux is 1.33 T, while expected average flux density in the core would be more like ~2.2 T; Probably as the surface is binding the air and the core is the reason --- and probably I would need to create an additional, slightly bigger, torus, and put the entire core in it to provide "stable conditions" for the core only.

UPDATE: added a tiny "skin" layer and now the inner core is "as expected", but now having issues with flux calculation.

Just Load State: B-field.pvsm for quick representation.
Field

raback:
How do you generate the mesh? I guess:
gmsh -3 mesh.geo 
ElmerGrid 14 2 mesh.msh -autoclean
I get some degenerated elements (~zero volume).

platise:
There is a Makefile which does everything, check your gmsh version (I use the latest from pypi), so:

$ gmsh -3 coreWire.geo

Info    : Running '/home/uros/python/bin/gmsh -3 coreWire.geo' [Gmsh 4.10.5, 1 node, max. 32 threads]
Info    : Started on Tue Oct 25 09:47:59 2022
Info    : Reading 'coreWire.geo'...
Info    : Done reading 'coreWire.geo'
Info    : Reading 'coreWire.geo.opt'...

$ ElmerGrid 14 2 coreWire.msh

Starting program Elmergrid
Output will be saved to file coreWire.

Elmergrid loading data:
-----------------------
Format chosen using the first line: $MeshFormat
Gmsh version is 4.1
Loading mesh in Gmsh format 4.1 from file coreWire.msh
Reading 6 entities in 0D
Reading 10 entities in 1D
Reading 10 entities in 2D
Reading 4 entities in 3D
Allocating for 37050 knots and 238743 elements.
Leading element dimension is 3
BC name for physical group 16 is: Sky
BC name for physical group 17 is: Wire End 1
BC name for physical group 18 is: Wire End 2
BC name for physical group 19 is: Wire Surface
BC name for physical group 20 is: Core Surface
BC name for physical group 21 is: Core Flux 1
BC name for physical group 22 is: Core Flux 2
Body name for physical group 23 is: Air
Body name for physical group 24 is: Wire
Body name for physical group 25 is: Core
Allocating lookup table for tags of size 13
Defined 6 0DIM entities with geometric tag range [3 8]
Defined 10 1DIM entities with geometric tag range [4 13]
Defined 10 2DIM entities with geometric tag range [4 13]
   Physical given tag range is [16 22]
Defined 4 3DIM entities with geometric tag range [2 5]
   Physical given tag range is [0 25]
   Physical read tag range is [23 25]
Physical tag offset for 1D is 22
Physical tag offset for 0D is 22
Geometric tag offset for 1D is 35
Geometric tag offset for 0D is 48
Reading 37050 nodes in 30 blocks.
Reading 238743 elements in 14 blocks.
Mapping mesh tag 4 to physical tag 19 in 2DIM
Mapping mesh tag 5 to physical tag 17 in 2DIM
Mapping mesh tag 6 to physical tag 18 in 2DIM
Mapping mesh tag 7 to physical tag 16 in 2DIM
Mapping mesh tag 8 to physical tag 16 in 2DIM
Mapping mesh tag 9 to physical tag 16 in 2DIM
Mapping mesh tag 10 to physical tag 20 in 2DIM
Mapping mesh tag 11 to physical tag 20 in 2DIM
Mapping mesh tag 12 to physical tag 21 in 2DIM
Mapping mesh tag 13 to physical tag 22 in 2DIM
Mapping mesh tag 2 to physical tag 24 in 3DIM
Mapping mesh tag 3 to physical tag 25 in 3DIM
Mapping mesh tag 4 to physical tag 25 in 3DIM
Mapping mesh tag 5 to physical tag 23 in 3DIM
Moving bulk elements to boundary elements
Assuming that boundaries names are already Ok!
Leading bulk elementtype is 504
Trailing bulk elementtype is 303
There are 13086 (out of 238743) lower dimensional elements.
Node 6962 belongs to maximum of 50 elements
Found 10664 side elements that have two parents.
Found correctly 13086 side elements.
Parent elements were reordered up to index 225657.
Moved 225657 elements (out of 238743) to new positions
Successfully read the mesh from the Gmsh input file.
Using physical numbering of entities

Elmergrid creating and manipulating meshes:
-------------------------------------------

Elmergrid saving data with method 2:
-------------------------------------
Saving mesh in ElmerSolver format to directory coreWire.
Reusing an existing directory
Saving 37050 coordinates to mesh.nodes.
Saving 225657 element topologies to mesh.elements.
Saving boundary elements to mesh.boundary.
Saving header info to mesh.header.
Saving names info to mesh.names.
Saving entities info to entities.sif.

Thank you for using Elmergrid!
Send bug reports and feature wishes to elmeradm@csc.fi

raback:
Ok, I split this as follows:

1) The *.F90 is now under fem/src/modules/MagnetoDynamics directory where it is automatically compiled. (I reintended it with our default two white spaces). 
2) New keywords in SOLVER.KEYWORDS
These both are in branch "PlatiseMerge" which I will merge if ok. 

3) The rest of the case into different repo where bigger cases are better suited:
https://github.com/ElmerCSC/elmer-elmag/commit/827dd1f81b9f7814c7a904caf81e0066b1a23d43
Here the test case can be further explained, pictures can be added. 

When time is ripe we could make a smaller consistency case under elmerfem/fem/tests.

platise:
Super inside the .geo file just maybe add  link to the https://www.elmerfem.org/forum/viewtopic.php?t=7820

I have added one comment/review regarding the missing Real at the Dipoles definition, you've mentioned above you had to add, otherwise it didn't work.

The rest I do not see any objections to merge to devel stream.

richb2k:
Regarding the missing Real, that is typical of a keyword that is not in the keywords file.  I ran the test with a nightly binary dated 2022-10-05, so that build pre-dates any commits.  Once the SOLVER.KEYWORDS file gets updated with your new keywords, then the error should go away.  On the other hand, adding 'Real' doesn't have a down side.

raback:
Yes this is how it goes. Btw, note that you can add some local keywords to your own SOLVER.KEYWORDS file (in the case directory) that will be appended to the other keywords.

raback:
I will close this as the stuff has essentially been added in the following:
https://github.com/ElmerCSC/elmer-elmag/tree/main/PlatiseFerroModel
https://github.com/ElmerCSC/elmerfem/commit/9e04183b76d359616cbf286ed84c0b730d117546

richb2k:
A rough draft of a consistency test using an elmergrid model can be found here:
http://www.elmerfem.org/forum/viewtopic.php?p=27321#p27321

platise:
Have added some short doc for quick usage: http://isotel.org/mixedsim/platiseferromodel

Of course background theory and usage can be "copied/contributed" to the solver manual.

Article is still on the TODO list, to give more insight on it.

PR #365, opened 2022-12-02
Refactor circuit element contributions

ettaka:
The point of this refactor is to make a function that could deal with all the element contributions no matter if they are type of boundary or volume. Then this function could be reused for impedance BCs.


ettaka:
Does this make sense to you @juharu ?
Maybe we should include also the matrix element creation also? That has to be done for the boundary elements also, for sure...

ettaka:
> Does this make sense to you @juharu ? Maybe we should include also the matrix element creation also? That has to be done for the boundary elements also, for sure...

I did it for the CircuitsUtils side also. I think it makes sense.

Happy independence day to all Finns!

juharu:
Looks OK to me - thanks! 
-J

PR #366, opened 2022-12-12
Add actions for mesh pieces to ElmerGUI Simulation section.

richb2k:
New ElmerGUI check box for 'Calculate Mesh Pieces', and new ElmerGUI text box to enter the number of 'Desired Mesh Pieces'.  
By default, if the check box is unchecked and the text box is empty, then nothing is added to the Simulation section of the sif file.  Checking the check box will add 'Calculate Mesh Pieces = Logical True'.  Likewise, if a number is entered into the text box, such as a '1', then 'Desired Mesh Pieces = Integer 1' will be added to the Simulation section.  If both the check box is checked and a number is added to the text box, then both statements will be added to the sif file.

### DIFF
--- a/ElmerGUI/Application/forms/generalsetup.ui
+++ b/ElmerGUI/Application/forms/generalsetup.ui
@@ -471,13 +471,34 @@
            </widget>
           </item>
           <item row="7" column="0">
+           <widget class="QCheckBox" name="calculateMeshPiecesCheck">
+            <property name="text">
+             <string>Calculate Mesh Pieces</string>
+            </property>
+           </widget>
+          </item>
+          <item row="7" column="2">
+           <widget class="QLabel" name="desiredMeshPiecesLabel">
+            <property name="text">
+             <string>Desired Mesh Pieces</string>
+            </property>
+           </widget>
+          </item>
+          <item row="7" column="3">
+           <widget class="QLineEdit" name="desiredMeshPiecesEdit">
+            <property name="text">
+             <string></string>
+            </property>
+           </widget>
+          </item>
+          <item row="8" column="0">
            <widget class="QLabel" name="simulationFreeTextLabel">
             <property name="text">
              <string>Free text</string>
             </property>
            </widget>
           </item>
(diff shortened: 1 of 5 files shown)


raback:
Nice work! I would just recommend adding the keywords "Calculate Mesh Pieces" and "Desired Mesh Pieces" to fem/src/SOLVER.KEYWORDS so that the type (Logical/Integer) does not need to be given. I know not all keywords are there but if they make the GUI they also earn the place in the SOLVER.KEYWORDS.

richb2k:
Okay, I'll work on adding both of those keywords.  I tried the other day, but without success, so it may take me a day or two to figure out what went wrong.

raback:
Adding the keyword unfortunately does not seem to update itself with "make install" or alike. For some reason you need a clean build. Or you can cheat by copying it manually. It is copied as is in style:
elmeruser@elmeruser-VirtualBox:~/elmerfem/fem/src$ cp SOLVER.KEYWORDS  /opt/elmer/share/elmersolver/lib/SOLVER.KEYWORDS

richb2k:
Working in Windows 10, I edited the solver.keywords file directly in C:\Program Files\Elmer 9.0-Release\share\elmersolver\lib, and the two added keywords are still not being recognized. 
ERROR:: LoadInputFile:  Unknown specifier: [true]
ERROR:: LoadInputFile:  In section: [simulation]
ERROR:: LoadInputFile:  For property name:[calculate mesh pieces]
I'll add the keywords in a separate branch, then if accepted, I'll wait for a windows binary build to try ElmerGUI again.

richb2k:
While looking for something else, I ran the test 'radiation2', and it failed with the same type of keyword error:
ERROR:: LoadInputFile:  Unknown specifier: [1.0e-10]
ERROR:: LoadInputFile:  In section: [solver 1]
ERROR:: LoadInputFile:  For property name:[minimum gebhart factor]
Do any other tests fail with this type of keyword error?

richb2k:
Update sif generator, removing Logical and Integer types from the sif commands.

PR #368, opened 2022-12-26
GUI icons update

t7saeki:
This PR is to replace Oxygen icons with Tango-based icons. New icons were made by using icons in Tango icon theme. Some icons were just taken from Tango icon theme and others were generated by combining some icons in Tango icon theme. Tango icon theme is in the public domain. So, there is no license issue like new version of Oxygen icons. 
Some of previous icons were too tiny to see in certain display settings. The new icons have multiple sizes (22x22, 32x32 and 48x48 for now) so as to solve such a problem.

PR #371, opened 2023-01-03
GUI context help

t7saeki:
This is a PR per some proposal on context help in GUI described in #368 by Rich_B
I have implemented some as listed below. Any feedback are welcome.

1.  "Get started" menu entry was added in Help menu to show a dialog. I have entered an example in the dialog. Please feel free to modify/delete/add code in MainWindow::getStartedSlot()   which starts from line 7291 of mainwindow.cpp. 
2. Blue "?" button was added in dialog for Equation, Solver setting, Material, etc. In windows OS, pressing small "?" button on title bar then clicking a widget is a standard manner to show context help, but I think people hardly be aware. So, I added a normal button in the right bottom corner of each dialog.  
3. To give users more chance to notice the existence of context help, I also experimentally put context help also into tooltip which shows up just by hovering mouse cursor on the widget for one or two seconds. I will eliminate this if you feel this is too much redundant.

Saeki


richb2k:
Very nice!  Thanks for the PR, looks good, but it will take a day or two to check it out, due to other priorities.  Rich.

richb2k:
Hello again.  The Get Started help menu item looks good.  I could add a little text to it later when time permits, but honestly it could be added right now.  Maybe break out the help menu item as a separate branch so it could be accepted in the near future?

The added button at the lower right is a great idea.  It would look and act the same in Linux as in Windows, correct?  Acting the same will help, the documentation then would be the same for both OSs.  Maybe label it as 'Whatis' instead of a question mark?

For the third item, hovering of the mouse seems to work well.  As you mentioned, perhaps it will be too much.  Maybe put it away for now, but be sure to keep a copy!

t7saeki:
Thank you for your comment, Rich! 
As for help menu, Yes please feel free to start a new branch and modify it.
As for the question mark button at the lower right corner, Yes it look and act the same in Linux and Windows. I have added 'Whatis' label in the button. 
As for hovering of mouse, I have commented it out so that we can revisit if necessary.

richb2k:
Hello,
The changes look good to me!  I compiled the branch on Windows and in the Elmer VM, everything seems to work as intended.  It might be an artifact of compiling a branch, but for my case in the Elmer VM, the 'Whatis' button works properly, but is blank, while in Windows it says 'Whatis'.  Rich.

richb2k:
Hello,
The 'Whatis' button works well and looks good, after installing binaries in both Windows and Elmer VM.

One more question, would it be possible to add the 'Whatis' button into the 'Setup' window?  Rich.

t7saeki:
Yes, it is possible to add 'Whatis' button into the 'Setup' window, but it will show nothing for now because no information is specified in elmerfem/ElmerGUI/Application/forms/generalsetup.ui. So, someone have to provide information to be shown for each item by editing the generalsetup.ui with Qt Designer (or directly editing with a text editor).

PR #372, opened 2023-01-03
GUI: Allow array parameter in text field

t7saeki:
This PR is to enable users to specify array parameters in text filed such like Save Coordinates(n,DIM) of SaveScalars solver.
For example, users can write in text filed like:
    

> (2,3) = 1.0 2.2 4.3 2.5 4.3 1.3

Then, generated sif code will be 

> Save Coordinates(2,3) = 1.0 2.2 4.3 2.5 4.3 1.3

This is simply done by avoiding adding "=" by SifGenerator when text filed starts with parenthesized array size followed by "=".


raback:
Nice little feature!

I have been thinking also whether it would be a good idea to always check for number of integers and reals for a vector. Array needs to know the dimensions but a vector could be checked for. It has been this way since the dawn but really it could be rather easily altered in ElmerSolver parser. One of the common mistakes is to give, e.g.

Target Boundaries(2) = 1 3 8

missing the third entry....

PR #375, opened 2023-01-12
VTK postprocessor improvements

t7saeki:
This PR is to improve some behavior of VTK post processor in GUI. I guess VTK postprocessor is not so popular now, but I still love it.  
- Drawn boundaries of two bodies in 2D feature edges such like FsiObstacleGUI tutorial sample
- Selectable color for Null field ( Added 'Color select' button next to field selection)
- Selectable color for mesh points, mesh edges and feature edges in Preferences window (Menu items to turn on/off mesh points, mesh edges and feature edges in View menu were replaced by checkboxes in Preference window)
- Smoother color mapping ( interpolation and enhanced color resolution)

Hope you like it.
Saeki


richb2k:
After failing to compile the new branch with VTK in Windows (it wanted qt6, boost, etc.), and after learning how to use Gparted to increase the Elmer VM from 20 GB to 30 GB, compiling the new branch in the Elmer VM went smoothly.  
FsiObstacleGUI definitely looks better with the outline of both bodies.  The check boxes in Preferences work nicely, one can turn on/off points and meshes easily.  Colors looked good, although I didn't spend much time testing color selection.
Looks good to me!

richb2k:
I'm also a fan of ElmerVTK, it is easy to use, perfect for new users of Elmer.  

With that said, it would be nice if one could obtain the ElmerVTK graphical results with the same number of clicks as required in Paraview.   For example, in Paraview, the display window opens and one selects a desired variable and the graphical display shows the results with a color bar.  In ElmerVTK, open the display window, deselect 'Surface', select 'Isocontours' for 2D or 'Isosurfaces' for 3D, then select a variable under Contour Control, followed by selecting the same variable under Color Control, click Apply.  Next, open Color Bar, then select a Color Map such as Isocontours, and click on Apply.  By my count, Paraview takes one click, while ElmerVTK requires eight clicks.  

Maybe adding some additional check boxes to Preferences would be sufficient, such as 'remember last selected tab' to remember 'Isocontours' instead of 'Surfaces', or 'start by opening the last selected tab' where it would go ahead and open the 'Isocontours' menu upon starting ElmerVTK.  

Just a suggestion!  Thanks,  Rich.

t7saeki:
Thank you for suggestion! It must definitely improve user experience. I tried writing some code, but found that it is a little complicated. So, I will separate it to another blanch.  - Saeki

richb2k:
Hello Saeki,

I'm sure it will be complicated, qt is not the easiest program to use.  Even if only one or two clicks can be removed, that would be an improvement.

Thanks, Rich.

t7saeki:
Thank you for your comment, Rich! I have improved a little to hold last selected filters (or tabs) until ElmerGUI get closed as long as file path of the post file is identical.  This will be useful when reloading updated post file by 'reload' button on ElmerVTK or 'ElmerVTK' button on ElmerGUI. 

Other two modifications in ElmerVTK: 
- Showing file path in ElmerVTK title bar
- Fixed bug which shows the last read post file when launching ElmerVTK without loading project

Regards,
Saeki

richb2k:
First try, compiling in Windows with the cmake flag WITH_VTK set to false, mainwindow.cpp, line 2625 fails.  Looks like it needs a '#ifdef EG_VTK/#endif' guard.  Compiled successfully in Elmer VM with cmake flag WITH_VTK set to true.

t7saeki:
I have added the '#ifdef EG_VTK/#endif' guard.  Thank you very much for reporting!

raback:
Great work again Saeki! And Rich thank you for doing the valuable work of testing and inspiring. Great international collaboration!

PR #382, opened 2023-02-08
Fix solver specific option in ElmerGUI, here are the before and after…

richb2k:
… cases

before:
Variable = Potential[Potential Re:1 Potential:1]
after:
Variable = Potential[Potential Re:1 Potential Im:1]

Based on the bug report in the forums:
[http://www.elmerfem.org/forum/viewtopic.php?p=27916#p27916](url)

### DIFF
--- a/ElmerGUI/Application/src/sifgenerator.cpp
+++ b/ElmerGUI/Application/src/sifgenerator.cpp
@@ -905,7 +905,7 @@ bool SifGenerator::parseSolverSpecificTab(DynamicEditor *solEditor, const QStrin
       if( entry.elem.attribute("Widget", "") != "Edit") continue;
 
       QLineEdit *l = static_cast<QLineEdit*>(entry.widget);
-      QString varName = l->text().trimmed();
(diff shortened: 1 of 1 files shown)


raback:
Nicely spotted! I saw the discussion on this topic and could not just get my head around how the "im" (and only that) is dropped out! This is one of those cases where the fix looks very minimal but I bet you had to hunt it down for quite a bit!? Great!

richb2k:
Since the issue was generated by Freecad, I knew it had to be an ElmerGUI issue.  Then, after a bit of research about Qt, trimmed, split, and simplified, followed by lot's of 'cout <<', it became easier to track down.   Also, the previous PR to fix issue #276 explained that using 'simplified' is better for Elmer than 'trimmed', due to removing extra spaces from sentences while still trimming trailing spaces.

Also, I checked other ElmerGUI edf's, such as Navier-Stokes solver, and there shouldn't be any other effects to existing projects.  This also fixed the same issue with vectorhelmholtz.xml.

PR #384, opened 2023-02-22
Elmerice weertman friction law

fgillet:
add Newton  for Non-Linear Weertman friction law in IncompressibleNSVec solver


tzwinger (commented):
Hi Fabien,
would you have a short test-case that uses that feature to be added to the main test suite under `fem/tests` ?

Sorry, seems to have been added during my last pull -  ignore (and as it is about Weertman sliding, it is OK to have them under `elmerice`-tree)

PR #385, opened 2023-02-24
Elmerice

tzwinger:


tzwinger:
This is the result (with a lot of resolved merge-conflicts) from the latest merge of first, the newton-linearization in the vectorized Naveier-Stokes solver to `elmerice` and then `devel` into `elmerice` - so potential for a lot of things that have gone wrong and I missed. Hence I made a pull request that other people can have a check before merging back to `devel`.

fgillet (approved):
difficult to review everything; seems ok for me;
Main remark is that changes that were introduced in Messages.F90 (Fatal) to call XIOS finalisation have been lost in the 2 previous merges from dev -> elmerice.
Can we put them back?

PR #389, opened 2023-03-22
Increase ElmerGrid max BCs limit

jvencels:
We have experienced a segmentation fault when running ElmerGrid for a case with >1000 BCs. Valgrind indicates Use of uninitialised value of size 8 @ LoadElmerInput (egnative.c:5468). Increasing the limit for max BCs solves the problem.


raback:
Thanx! Seems reasonable.

richb2k:
Looks like increasing the max boundaries and max bcs from 1000 each to 5000 each causes ElmerGrid to crash on Windows 10.  PR #395 has been submitted to lower the max from 5000 back down to 4000.  This change allows ElmerGrid to run properly again.  Hopefully the lower values will still be high enough to prevent the seg fault from happening.

PR #391, opened 2023-04-06
Fix for Output var name longer than Var name

jvencels:
Fix for the case when the Output variable name is longer than the Variable name.

Example:

Simulation
...
   Output variable 4 = "relative permeability re e"
...
End

Causes the following error
SingleSolver: Solver Equation string is: post process data SaveResult: ----------------------------------------- SaveResult: Saving results to file: mesh/h1.result At line 3779 of file /home/vencels/Projects/Elmer/elmerfem-orig/fem/src/ModelDescription.F90 Fortran runtime error: Substring out of bounds: upper bound (26) of 'var' exceeds string length (19)


raback:
Thanx for the fix Juris!

PR #395, opened 2023-05-02
Windows version of ElmerGrid would crash upon trying to save the mesh.

richb2k:
Lowering the Maxboundaries and Maxbcx from 5000 to 4000 allows ElmerGrid to run to completion.


richb2k:
Here is the bug report on the forum:

[http://www.elmerfem.org/forum/viewtopic.php?t=7978](url)

Apparently setting the max BCs to 5000 worked on Linux, but not on Windows.  This pull request lowered the max BCs to 4000, which seems to work on both Windows and Linux.

Refer also to Issue #467

PR #403, opened 2023-06-08
Add test case output files to `.gitignore`

### LINKED ISSUE #402: Add manually run test case output files to `.gitignore`
If one runs a test case from _fem/tests_ without `ctest`, the resulting converted mesh files and _TEST.PASSED_ etc. will be visible in source control. I think these should be added to the _.gitignore_ file.

nikosavola:
Drafted until checked if this contains all possible output files.

Closes #402

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,15 @@ fem/tests/CircuitsAndDynamics
 slurm*out
 ElmerGUI/CMakeLists.txt
 
+# Test-generated files
+fem/tests/**/*.boundary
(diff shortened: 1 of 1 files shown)


raback (commented):
Hi, Unfortunately we cannot ignore mesh.* files since quite many tests only includes these. I understand that many come as .msh or .grd filed that are parsed but we cannot ignore the native format. However, we can ignore TEST.PASSED and "results". I just committed those separately.

PR #408, opened 2023-06-19
Minor update to matc.  Allows 'exit' or 'quit' to work as expected when running from the command line.

richb2k:
Compiled and tested on Windows 10 and on Ubuntu 20 in the Elmer VM.

### DIFF
--- a/matc/src/main.c
+++ b/matc/src/main.c
@@ -74,9 +74,9 @@ int main( int argc, char **argv )
       
 /* kludge to enable exit. */
 #if defined(WIN32) || defined(MINGW32)
-      if( stricmp(str,"exit") == 0  || stricmp(str,"quit") == 0 )
+      if( stricmp(str,"exit\n") == 0  || stricmp(str,"quit\n") == 0 )
 #else
-      if( strcasecmp(str,"exit") == 0  || strcasecmp(str,"quit") == 0 )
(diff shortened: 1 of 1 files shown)


richb2k:
This update is a change to the executable version of matc, that can be called from the command line.  This change will not effect calls to mtc_domath().

juharu:
Merged now, thanks Rich!

richb2k:
You are welcome.  I have a question about matc and elmergrid.  I see there is an option to compile elmergrid with matc, and in fact it works.  Is there a historical reason to not compile elmergrid with matc for production?
Rich.

raback:
There is really no reason not to compile ElmerGrid with MATC. Could be useful for some users.

richb2k:
I guess the hardest part would be to change cmake to build them together.

richb2k:
So, I used building of matc with a shared library as a model, as it is currently implemented.  I made some changes to cmake, and now elmergrid will build with matc activated.  As a side product, libelmergrid.dll is created.  Rather than adding a PR for something experimental, take a look at this github repo that has the changes:  [https://github.com/richb2k/elmerfem](url)  If someone could check that building and packaging of elmerfem works properly, then I'd be happy to add a PR.  Thanks,  Rich.
Edit:  the changes compile and are working in Windows 10.  Elmergrid can process matc statements with the change in cmake.  Not sure how the changes will work with packaging, such as storing the dll properly in bin.

Issue #414, opened 2023-08-11
ElmerGUI fails to build when using OCC


GenericNerdyUsername:
I'm trying to migrate a package repository away from using [OCE](https://github.com/tpaviot/oce) as it is no longer maintained, but compiling with OCC and `-DWITH_ELMERGUI:BOOLEAN=TRUE` on commit `d1d4983` causes the compile to fail with the error
```
/build/source/ElmerGUI/Application/cad/cadview.cpp:365:54: error: 'class Poly_Triangulation' has no member named 'Nodes'; did you mean 'Node'?
  365 |     const TColgp_Array1OfPnt &Nodes = Triangulation->Nodes();
      |                                                      ^~~~~
      |                                                      Node

```

Issue #407, opened 2023-06-17
cadview Triangulation->Nodes() gives error


gianlucapettinello:
As written in the title when compiling elmerfem I get an error in cadview.cpp because Triangulation->Nodes() is not used any more I think bay opencascade.
Node() should be used. I'm trying to set Nodes.Lower() to 1 and Nodes.Upper() to nofNodes.
Thanks for checking
Gianluca

Issue #304, opened 2021-11-15
Please make a release because 9.0 isn't compatible with opencascade-7.6.0


yurivict:

yurivict:
```
/disk-samsung/freebsd-ports/science/elmerfem/work/elmerfem-release-9.0-685-g33126a3f/ElmerGUI/Application/cad/cadview.cpp:362:61: warning: 'Triangles' is deprecated: Deprecated method, Triangle() should be used instead [-Wdeprecated-declarations]
    const Poly_Array1OfTriangle &Triangles = Triangulation->Triangles();
                                                            ^
/usr/local/include/OpenCASCADE/Poly_Triangulation.hxx:317:3: note: 'Triangles' has been explicitly marked deprecated here
  Standard_DEPRECATED("Deprecated method, Triangle() should be used instead")
  ^
/usr/local/include/OpenCASCADE/Standard_Macro.hxx:154:54: note: expanded from macro 'Standard_DEPRECATED'
  #define Standard_DEPRECATED(theMsg) __attribute__((deprecated(theMsg)))
                                                     ^
/disk-samsung/freebsd-ports/science/elmerfem/work/elmerfem-release-9.0-685-g33126a3f/ElmerGUI/Application/cad/cadview.cpp:363:54: error: no member named 'Nodes' in 'Poly_Triangulation'
    const TColgp_Array1OfPnt &Nodes = Triangulation->Nodes();
                                      ~~~~~~~~~~~~~~~^
/disk-samsung/freebsd-ports/science/elmerfem/work/elmerfem-release-9.0-685-g33126a3f/ElmerGUI/Application/cad/cadview.cpp:832:8: warning: expression result unused [-Wunused-value]
  for (expFace; expFace.More(); expFace.Next()) {
       ^~~~~~~
/disk-samsung/freebsd-ports/science/elmerfem/work/elmerfem-release-9.0-685-g33126a3f/ElmerGUI/Application/cad/cadview.cpp:840:10: warning: expression result unused [-Wunused-value]
    for (expEdge; expEdge.More(); expEdge.Next()) {
         ^~~~~~~
3 warnings and 1 error generated.
```

opencascade-7.6.0
OS: FreeBSD 13

PR #416, opened 2023-08-14
GUI library upgrade

t7saeki:
This PR is for upgrading libraries for GUI.

- Option for Qt6
  AFAIK Qt5 is alraedy End of (normal) Support and MSYS2 already adopted Qt6 in VTK9.2 package. On the other hand, VTK9 package in most Linux distro still remains in Qt5. So option for Qt6 is provided to users by setting WITH_QT6 TRUE.

- OCCT version problem 
  Fix for issue #304, #407 and #414. (Note that OCE is still an option)


raback:
Great to have Qt6!

Issue #418, opened 2023-08-18
sifgenerator.cpp: error: qualified-id in declaration before '(' token


barracuda156:
```
[ 60%] Building CXX object ElmerGUI/Application/CMakeFiles/ElmerGUI.dir/src/solverparameters.cpp.o
/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_science_elmerfem/elmerfem/work/elmerfem-4f81fef4be6be8815f111b13046cd0758adec692/ElmerGUI/Application/src/sifgenerator.cpp: In member function 'void SifGenerator::handleLineEdit(const QDomElement&, QWidget*)':
/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_science_elmerfem/elmerfem/work/elmerfem-4f81fef4be6be8815f111b13046cd0758adec692/ElmerGUI/Application/src/sifgenerator.cpp:1312:34: error: qualified-id in declaration before '(' token
 1312 | void SifGenerator::handleTextEdit(const QDomElement &elem, QWidget *widget)
      |                                  ^
/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_science_elmerfem/elmerfem/work/elmerfem-4f81fef4be6be8815f111b13046cd0758adec692/ElmerGUI/Application/src/sifgenerator.cpp:1319:34: error: qualified-id in declaration before '(' token
 1319 | void SifGenerator::handleComboBox(const QDomElement &elem, QWidget *widget)
      |                                  ^
/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_science_elmerfem/elmerfem/work/elmerfem-4f81fef4be6be8815f111b13046cd0758adec692/ElmerGUI/Application/src/sifgenerator.cpp:1332:34: error: qualified-id in declaration before '(' token
 1332 | void SifGenerator::handleCheckBox(const QDomElement &elem, QWidget *widget)
      |                                  ^
/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_science_elmerfem/elmerfem/work/elmerfem-4f81fef4be6be8815f111b13046cd0758adec692/ElmerGUI/Application/src/sifgenerator.cpp:1351:2: error: expected '}' at end of input
 1351 | }
      |  ^
[...]

PR #420, opened 2023-08-21
Fix for #418

t7saeki:
My previous PR #416 caused the issue #418 which happens when building with Qt4. This is a fix for the issue.


barracuda156:
@t7saeki @raback Thank you!

PR #428, opened 2023-12-07
Implement post processing for homogenization

ettaka:
Here we implement post processing for homogenization model. New field, called the "proximity loss", is created and it contains the proximity losses caused by the homogenization parameter `nu`. Skin losses are added to the existing joule losses.

@juharu could you check if this is OK. I ran tests, they seem fine.

FYI @raback


ettaka:
May I merge this? @raback @juharu

PR #430, opened 2023-12-07
3D calving algorithm

iwheel:
The 3D calving algorithm from the St Andrews repository with the develop branched merged in. I had to make some changes to fem/src/MeshRemeshing.F90 following the merge that may affect other uses of mmg although it seemed ok on testing. I tried to prevent this by adding in the MultipleInputs options as the calving will use multiple Hmin/Hausd inputs that aren't present for other uses. Anyway it may be best to discuss in the future so I can align the calving algo with the global mmg uses. The other change of note is the timestepping in ElmerSolver.F90 as the calving algo modifies its own timesteps. This doesn't other uses.

Happy to discuss further as there are a lot of changes.

Iain


raback:
This is huge! You have done a lot! Maybe we should set up a zoom meeting to discuss the procedure.

raback:
Merging happening step-by-step in branch calving_meshadapt. I close this one.

PR #431, opened 2023-12-09
Harmonize ElmerGrid, fix debug io

richb2k:
line length limitation, and add printf statements about using MATC when not compiled,

### DIFF
--- a/ElmerGUI/Application/plugins/egconvert.cpp
+++ b/ElmerGUI/Application/plugins/egconvert.cpp
@@ -97,7 +97,7 @@ static int GetrowDouble(char *line1,FILE *io)
 
   for(i=0;i<MAXLINESIZE;i++) { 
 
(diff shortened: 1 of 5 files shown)


richb2k:
Added date compiled to ElmerGUI console output, to be consistent with ElmerSolver and ElmerGrid.

richb2k:
Add diagnostic output to the ElmerGUI solver log window, which is useful for parallel runs.  Rewrite the error message that is output when the MPI executable is not found.

raback:
Hi Rich, Sorry for taking time. I had a look while doing something else and merged the fix. I had already done the harmonization part so I just picked the one changes file and will not use the PR to merge. Thanx for the fix!

richb2k:
Hello Peter,  thanks for accepting the fix.  Rich.

Issue #451, opened 2024-02-28
Running multiple Elmer configurations at once sharing the same detached XIOS instance


lucas-ige:
Hello,

An execution ID for Elmer is required when using XIOS (cf. xios_initialize). In an MPI run that calls multiple programs (MPMD), this ID is used to group MPI processes by program, so that each program can implement internal parallelization without interfering with other programs. As of now, this ID is hard-coded (variable xios_id in fem/src/Types.F90, with value "elmerice"). It is therefore not possible to run different configurations of Elmer in detached XIOS mode simultaneously in a single MPI run because they all have the same ID while they should have different IDs.

I am hereby suggesting a commit to make this possible. With this commit, the user can specify the ID at run-time using the -exec-id command line argument. For example:

`mpirun -np 10 ./ElmerSolver_mpi greenland.sif -exec-id greenland : -np 1 ./xios
`

For **backward compatibility**, -exec-id can be omitted, and it defaults to "elmerice".

Internally, the variable xios_id has been renamed to ExecID because it can be used outside of XIOS (for example: external couplers like OASIS may rely on such an ID).

This has applications in climate modelling where multiple configurations of Elmer may need to be run at once, sharing a single XIOS server (for example: one for Greenland and one for Antarctica). This could also be useful for ensemble modelling.

Please note that the context ID in Elmer's XIOS configuration files should match the value of ExecID.

I updated the XIOSOutputSolver's documentation accordingly. There might be other places where the documentation should also be updated?

I am looking forward to hearing your thoughts regarding this suggestion.

Cheers,

Lucas Bastien

lucas-ige:
This is the [pull request](https://github.com/ElmerCSC/elmerfem/pull/452) I am referring to.

tzwinger:
I merged - please test and let me know whether we can close this issue

lucas-ige:
Hello, thank you for the merge.

I tested the merged code: it works as intended (I ran one simulation with two icesheets at once and another simulation with a single icesheet to test the default backward-compatible behavior).

This issue can be closed, as far as I am concerned. Quick question though: I documented this _-exec-id_ feature in [the documentation of the XIOSOutputSolver](https://github.com/ElmerCSC/elmerfem/blob/elmerice/elmerice/Solvers/Documentation/XIOSOutputSolver.md); should it also be documented somewhere else? If yes, I'm happy to do it. If no, then feel free to close this issue.

PR #452, opened 2024-02-28
Implement dynamic execution ID (replaces xios_id).

lucas-ige:
Context: an execution ID is required when using XIOS (cf. xios_initialize). Before this commit, this ID was hard-coded (variable xios_id in fem/src/Types.F90, with value "elmerice"). It was therefore not possible to run different configurations of Elmer in detached XIOS mode simultaneously in a single MPI run (each configuration uses this ID to identify its own MPI instances among all the instances).

This becomes possible with this commit: the user can now specify the execution ID at run-time using the -exec-id command line argument. For example:

mpirun -np 10 ElmerSolver_mpi greenland.sif -exec-id greenland : -np 1 xios

For backward compatibility, -exec-id can be omitted, and it defaults to "elmerice".

Internally, the variable xios_id has been renamed to ExecID because it can be used outside of XIOS (for example: external couplers like OASIS may rely on such an ID).

This has applications in climate modelling where multiple configurations of Elmer may need to be run at once, sharing a single XIOS server (for example: one for Greenland and one for Antarctica). This could also be useful for ensemble modelling.

Please note that the context ID in Elmer's XIOS configuration files should match the value of ExecID.

### DIFF
--- a/elmerice/Solvers/Documentation/XIOSOutputSolver.md
+++ b/elmerice/Solvers/Documentation/XIOSOutputSolver.md
@@ -1,15 +1,15 @@
 # Solver XIOSOutputSolver
 ## General Information
-- **Solver  Fortran File:** XIOSOutputSolver.F90   
-- **Solver/User Function Name:** XIOSOutputSolver  
-- **Required Output Variable(s):** None  
-- **Required Input Variable(s):** None  
-- **Optional Output Variable(s):** None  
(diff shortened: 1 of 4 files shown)


lucas-ige:
This pull request is linked to [this issue](https://github.com/ElmerCSC/elmerfem/issues/451).

PR #463, opened 2024-04-16
Fix saveline

fgillet:
Fix for two issues in SaveLine;
First, do not use DGVar output mode if there is no DG variable to save; 
Second, fix issues with column order not preserved if one of the variable is not defined on a given boundary.


raback:
Looks legit. Thanx for fixing!

Issue #460, opened 2024-04-03
CoilSolver Crappy potentials for multi-turn circuit case


jvencels:
@ettaka 
The geometry has two simple turns.
![image](https://github.com/ElmerCSC/elmerfem/assets/5177281/328a75e1-1e89-4dbd-8767-28cd769e7ba5)

oneCoil.sif - we connect only one turn to circuits, another one is modeled as air. This works.
series.sif - we connect both turns in series using circuits and get an error.
[CoilSolver_Circuits.zip](https://github.com/ElmerCSC/elmerfem/files/14847866/CoilSolver_Circuits.zip)

These are the difference in cases:
On the left sif where turns connected in series, on the right - only one turn
![image](https://github.com/ElmerCSC/elmerfem/assets/5177281/4f2540a6-55b1-4021-8c5d-808a7aaee7d1)
Further, there are differences in circuit equations - one component and two components. The circuit setup should be ok as we use it all the time.

jvencels:
![vMRuS2jmJqSPavmG](https://github.com/ElmerCSC/elmerfem/assets/5177281/fb95346c-e941-425d-8711-910cdbcac8df)
Comparison. On the left it runs, on the right it fails.

ettaka:
I looked around a bit. Didn't find the problem yet, but I cleaned the case up so it will be easier to narrow the problem down:
[only_coil_solver.zip](https://github.com/ElmerCSC/elmerfem/files/14913723/only_coil_solver.zip)

jvencels:
I tried running the case, getting convergence problem:

```
     500 0.4154E+10
     501 0.4154E+10
ERROR:: IterSolve: Numerical Error: Too many iterations were needed.
STOP 1
```

@ettaka Can you reproduce "crappy potentials" error with cleansed case?

ettaka:
@jvencels yes, I did reproduce it also. I must have uploaded some other try. I will upload another

ettaka:
This gives me the error you mention:
[only_coil_solver.zip](https://github.com/ElmerCSC/elmerfem/files/14933276/only_coil_solver.zip)

ettaka:
I see that the `coil current` is fine, but the `coil current e` is terrible
![image](https://github.com/ElmerCSC/elmerfem/assets/975622/2aa0b6c3-9830-42be-9cbc-2a23c5fcfac7)

[...]

PR #468, opened 2024-05-28
Fix Homogenization 3D

ettaka:
* Don't require homogenization for all coils
* Add missing nu component (homog)

  By convention, we are using the 2D approximation in 2D stranded homogenization parameters: in local coordinates they are xx and yy. But, the third one (zz) has been missing, causing some strange tensors as output. In this commit the third is added. Note that now the third has to be declared in the sif:

  ```sif Component: Nu 33 Component: Nu 33 im ```

  Had to touch these: `fem/tests/circuits_harmonic_homogenization_coil_solver/sif/2241.sif` `fem/tests/circuits_harmonic_stranded_homogenization/sif/2241.sif`

TODO:
  * we need to hack coil solver such that we don't have the problem shown in: https://github.com/ElmerCSC/elmerfem/issues/460#issuecomment-2059231516 This should be fixed soon! @raback


raback:
Thanx! Have to look at the crappy potential issue.

PR #473, opened 2024-06-13
Cleanup

3x380V:
No functional changes comes with this pull request. It is just cleanup born as a byproduct of looking around code base.


3x380V:
Rebased on current HEAD. Those rebases will occur on semi-regular basis to make merging easier (in hope anyone ever notices this pull request)

raback:
Thanx for polishing things up! One could perhaps have the Hutiter documentation under "elmerfem-manuals" repo which includes all other docs as well. But maybe not worth the effort as the end users don't much need to read these docs. Certainly the PostScipt format better make way for pdf.

PR #480, opened 2024-07-12
Build rules: Increase `cmake_minimum_required` to 3.10.

mmuetzel:
This is possibly a first step in a series of changes that update the build system to make it a bit easier to build ElmerFEM with current build tools and dependencies.
Are you interested in such changes?

The latest versions of CMake are no longer capable of running in compatibility mode for CMake 1.0.2. This leads to a lot of warnings being displayed during configuration.

Most distributions have updated to newer versions. 
E.g., some widely used distributions with a long support period and their CMake versions:
Debian 10: CMake 3.13.4
OpenSUSE Leap 15.5: CMake 3.20.4
RHEL 8: CMake 3.20.2
Ubuntu 18.04: CMake 3.10.6

CMake 3.10 was released in 2017. It is distributed by most LTS distributions that are close to end of life (or already reached it).
It is also the first version to implement OpenMP components (which will be relevant in a later change).

Increase `cmake_minimum_required` to 3.10.
Set the minimum CMake version *before* creating a project where applicable.


raback:
Seems legit. Even RedHat is has never cmake.

PR #481, opened 2024-07-12
Remove copies of FindMPI and FindOpenMP CMake modules.

mmuetzel:
These modules have been copied from a (very) old version of CMake (probably a long time ago). The minimum CMake required version has been bumped to 3.10 recently which include these modules.

Additionally, the existing copy of the (old) FindOpenMP CMake module is incompatible with the CMake import targets used by newer versions of SuiteSparse libraries (CHOLMOD and UMFPACK). (That is only important when trying to use non-bundled versions of those libraries.)

Just remove the two modules that are no longer needed plus their dependencies (since they are supplied by CMake itself) and only cause issues.


raback:
Indeed, this seems to work without both with MPI and OpenMP at least on my Ubuntu. Thanx for contributing! The built system rarely gets the needed attention. Just running on other platforms and will merge if tests pass.

PR #483, opened 2024-07-12
Simplify configure checks for OpenMP by relying on upstream CMake.

mmuetzel:
Upstream CMake already check which version of OpenMP the used compilers support. Use that information to simplify the configuration a bit.

Also fix what is probably a typo in the build rules.


mmuetzel:
Rebased on a current head without any changes to check if the CI triggers.

PR #484, opened 2024-07-13
Add CI rules for GitHub hosted Ubuntu runners.

mmuetzel:
These rules are heavily inspired by the existing CI scripts (apart from Zoltan that didn't build for me).

Building with LLVM Clang and libomp is currently failing with a linking error. That might be caused by how the CMake targets link to OpenMP in the build rules. (But I haven't looked into it any further. Might try to dig into that later.)
I left that part of the build matrix commented out for the time being.

Additionally, some tests were failing for me. I don't know whether those test failures are expected or if that's because of a configuration error in the proposed CI rules.

I was looking into using `ccache` to potentially speed up compilation time. But there are three main reasons why I opted against that:
* A large portion of the code of ElmerFEM is written in Fortran for which current versions of `ccache` don't have support.
* Even for the remaining parts of the code (written in C or C++) a large portion can't be cached because `config.h` includes the Git revision. So, almost every compilation unit changes on every push which would invalidate the compiler cache.
* The CI time is dominated by the test suite anyway. (Don't get me wrong: Having good coverage is actually a good thing.)

The proposed rules would trigger on each change of a PR and whenever a change is pushed to the repository. Additionally, it would be possible to trigger the CI manually for maintainers of this repository. The same rules are run in either case. Is that agreeable?

Are you interested in having CI running on the GitHub-hosted runners? If you are, GitHub also hosts runners with macOS and Windows that are free (as in free beer) to use for open source projects. So, once the CI rules are sorted out for the Ubuntu runners, we could look into adding rules for the other OSs if that is something you are interested in.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,97 @@
+name: build
+on:
+  workflow_dispatch:
+  push:
+  pull_request:
+
+concurrency: ci-${{ github.ref }}
+
+jobs:
+
+  ubuntu:
+    # For available GitHub-hosted runners, see:
+    # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners
+    runs-on: ubuntu-latest
+
+    name: ubuntu (${{ matrix.compiler }})
+
+    strategy:
+      # Allow other runners in the matrix to continue if some fail
+      fail-fast: false
+
+      matrix:
+        compiler: [gcc]
+        include:
+          - compiler: gcc
+            compiler-pkgs: "g++ gcc"
+            cc: "gcc"
+            cxx: "g++"
+          #- compiler: clang
+          #  compiler-pkgs: "clang libomp-dev"
+          #  cc: "clang"
+          #  cxx: "clang++"
+
+    env:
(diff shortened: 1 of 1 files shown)


mmuetzel:
Looks like actions aren't run (yet) for this repository. Maybe, a maintainer would need to enable them?

For an idea of how the result of running this action might look like: 
https://github.com/mmuetzel/elmerfem/actions/runs/9911987087/job/27385837816

raback:
CI would certainly be useful! We had Travis working long time ago but then it was no longer free.  

The tests do consume a lot of time. For that reason there are flags. "quick" for ~<1s and "slow" for ~>10s. To run just quick tests "ctest -L quick -j4", for example.

raback:
This is great! Thank you very much! I accepted the PR and activated the Actions and everything worked out-of-box! It now runs just the quick tests taking about half on hour.

Could there by multiple workflows such that they would be launched with different rules? Running all tests less frequently would be nice. Also, activating the "elmerice" capabilities/tests will bloat the testing. If testing takes ~2h it is often not that useful for intra-day testing.

PR #485, opened 2024-07-13
Add CI rule that triggers once a week with the full test suite. 

mmuetzel:
Currently, this is basically a copy of the build rules that trigger on pull requests or pushes. The only difference is that the trigger is on a weekly schedule and the full set of tests are run.

Mid-term, more tests could be run for different configurations on PRs or pushes (e.g., build tests with bundled or external dependencies, with some enabled or disabled features, with different compilers, on different platforms, ...) while the full test suite is only run with one or a different subset of configurations.

That will be easier if we start with two sets of (currently basically identical) rules and let them spread in different directions from here without having to care about the other rules.

I don't know what "activating the "elmerice" capabilities/tests" means. So, I didn't touch anything in the existing configuration.
Could you please let me know if something should be changed?

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -55,7 +55,6 @@ jobs:
           cmake \
             -DCMAKE_BUILD_TYPE="Release" \
             -DCMAKE_INSTALL_PREFIX="${GITHUB_WORKSPACE}/usr" \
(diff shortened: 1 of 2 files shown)


mmuetzel:
Modifying the schedule to run more or less often should be quite straight-forward. Please, let me know if you prefer to do that as part of this PR.

Issue #486, opened 2024-07-13
Building with LLVM Clang and OpenMP fails on Ubuntu 24.04


mmuetzel:
When trying to build with LLVM Clang 18.1.3 on Ubuntu 24.04, linking `ViewFactors` fails with the following error when EulerFEM is configured to build with OpenMP (`CC=clang CXX=clang++ cmake -DWITH_OpenMP=ON ..`:
```
/usr/bin/ld: view3d/libview3d.a(ViewFactors.c.o): undefined reference to symbol '__kmpc_dispatch_next_4@@VERSION'
/usr/bin/ld: /lib/x86_64-linux-gnu/libomp.so.5: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```

I don't know what causes that error or how to resolve it.

Building works after having configured with `-DWITH_OpenMP=OFF` (the default if `WITH_OpenMP` isn't set manually).

Is that a known issue?

tzwinger:
Did I get it wrong, but are you still using `gfortran` for the Fortran part of Elmer? Simply, it works for me using `clang` and `clang++`, but as soon as I replace `gfortran` with `flang`, I get into all sorts of troubles. For CSC it would be very interesting to get it working with the complete LLVM compiler suite, as LLVM is important for our supercomputer LUMI, which has AMD hardware.

mmuetzel:
Yes, this is still with `gfortran`. I only used `clang` and `clang++` instead of `gcc` and `g++`.
The reason for that error might be that some dependencies (OpenBLAS?) were built with `libgomp` on Ubuntu. And linking different OpenMP implementations into the same application causes errors.

When it comes to using LLVM Flang: MSYS2 (Windows MinGW) has the CLANG64 environment that is based on a LLVM toolchain (`clang`, `clang++`, `flang`, `lld`, `libomp`, ...):
https://www.msys2.org/docs/environments/

I'd like to add some CI runners for MSYS2 (incl. a runner using their CLANG64 environment) once the CMake build rules are ready for Windows. That would hopefully cover at least parts of the build configuration on your supercomputer.

ghostforest:
I got a similar error. The linker cant find openMP or something. I suggest you check here: https://github.com/ElmerCSC/elmerfem/issues/503#issuecomment-2248193190
[...]

PR #487, opened 2024-07-13
Add runner that builds with Clang to CI.

mmuetzel:
Building with LLVM Clang and libomp currently fails (see #486).

Add configurations to the build matrix that build with LLVM Clang without OpenMP. Also add a runner that builds with GCC without OpenMP.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -13,23 +13,38 @@ jobs:
     # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners
     runs-on: ubuntu-latest
 
-    name: ubuntu (${{ matrix.compiler }})
+    name: ubuntu (${{ matrix.compiler }} ${{ matrix.openmp }} OpenMP)
 
     strategy:
       # Allow other runners in the matrix to continue if some fail
       fail-fast: false
 
       matrix:
-        compiler: [gcc]
+        compiler: [gcc, clang]
+        openmp: [with, without]
         include:
           - compiler: gcc
             compiler-pkgs: "g++ gcc"
             cc: "gcc"
             cxx: "g++"
-          #- compiler: clang
-          #  compiler-pkgs: "clang libomp-dev"
-          #  cc: "clang"
-          #  cxx: "clang++"
+            openmp: with
+            openmp-cmake-flags: "-WITH_OpenMP=ON"
+          - compiler: gcc
+            compiler-pkgs: "g++ gcc"
+            cc: "gcc"
+            cxx: "g++"
+            openmp-cmake-flags: "-WITH_OpenMP=OFF"
+          - compiler: clang
+            compiler-pkgs: "clang"
+            cc: "clang"
+            cxx: "clang++"
+            openmp: with
+            openmp-cmake-flags: "-WITH_OpenMP=ON -DOpenMP_C_FLAGS=-fopenmp=libgomp -DOpenMP_CXX_FLAGS=-fopenmp=libgomp"
+          - compiler: clang
+            compiler-pkgs: "clang"
+            cc: "clang"
+            cxx: "clang++"
+            openmp-cmake-flags: "-WITH_OpenMP=OFF"
 
     env:
       CC: ${{ matrix.cc }}
@@ -59,7 +74,7 @@ jobs:
             -DCMAKE_BUILD_TYPE="Release" \
             -DCMAKE_INSTALL_PREFIX="${GITHUB_WORKSPACE}/usr" \
             -DBLA_VENDOR="OpenBLAS" \
-            -DWITH_OpenMP=ON \
+            ${{ matrix.openmp-cmake-flags }} \
             -DWITH_LUA=ON \
             -DWITH_Zoltan=ON \
             -DWITH_Mumps=ON \


mmuetzel:
It looks like a few tests fail with Clang that passed with GCC:
```
The following tests FAILED:
	118 - CurvedBoundaryP (Failed)
	136 - DisContBoundaryFullAngle (Failed)
	219 - FixTangentVelo (Failed)
	255 - IncreaseOrderTet (Failed)
```

mmuetzel:
As the motivation why I think testing with LLVM Clang on Linux could be useful:
LLVM Clang and Apple Clang are not identical. But they are pretty similar.
Many developers (including me) don't have physical access to a macOS device.
Sorting out potential issues with LLVM Clang on Linux often leads most of the way towards successfully building a project with Apple Clang on macOS.

mmuetzel:
The default log file is not very useful to try and understand what is happening in these failing tests. If I understand correctly, stderr and stdout are redirected to test-specific files. That makes it difficult to display the output in automated tests.

Would it be ok to no longer redirect the output and let `ctest` do "its thing" instead?

raback:
Looks great! It is nice to be able to test several builds at the same time. 

Previously we seem to have run the tests with:
ctest -L quick => 'CTEST_OUTPUT_ON_FAILURE=1 ctest -L quick' 
To allow output in CI.

mmuetzel:
If I understand their documentation correctly, `--output-on-failure` should have the same effect as `CTEST_OUTPUT_ON_FAILURE=1`:
https://cmake.org/cmake/help/latest/manual/ctest.1.html#id18
> `--output-on-failure`
Output anything outputted by the test program if the test should fail. This option can also be enabled by setting the [`CTEST_OUTPUT_ON_FAILURE`](https://cmake.org/cmake/help/latest/envvar/CTEST_OUTPUT_ON_FAILURE.html#envvar:CTEST_OUTPUT_ON_FAILURE) environment variable

To be honest, the output that is currently shown for the failing tests might be enough for someone more familiar with ElmerFEM to understand what is happening. It's just that I didn't understand what it means.
E.g., for `CurvedBoundaryP`, the test output is:
```
     Start 118: CurvedBoundaryP
  1/4 Test #118: CurvedBoundaryP ..................***Failed    2.55 sec
  -- BINARY_DIR = /home/runner/work/elmerfem/elmerfem/build
  -- Extra library directories 
  CMake Error at /home/runner/work/elmerfem/elmerfem/cmake/Modules/test_macros.cmake:157 (FILE):
    FILE failed to open for reading (No such file or directory):
  
      /home/runner/work/elmerfem/elmerfem/build/fem/tests/CurvedBoundaryP/TEST.PASSED
    /home/runner/work/elmerfem/elmerfem/fem/tests/CurvedBoundaryP/runtest.cmake:3 (RUN_ELMER_TEST)
```

And the corresponding output in the test log is:
```
Start testing: Jul 13 21:44 UTC
  ----------------------------------------------------------
  118/870 Testing: CurvedBoundaryP
  118/870 Test: CurvedBoundaryP
  Command: "/usr/local/bin/cmake" "-DCMAKE_MODULE_PATH=/home/runner/work/elmerfem/elmerfem/cmake/Modules" "/usr/local/share/cmake-3.30/Modules" "-DELMERGRID_BIN=/home/runner/work/elmerfem/elmerfem/build/elmergrid/src/ElmerGrid" "-DELMERSOLVER_BIN=/home/runner/work/elmerfem/elmerfem/build/fem/src/ElmerSolver_mpi" "-DFINDNORM_BIN=/home/runner/work/elmerfem/elmerfem/build/fem/tests/findnorm" "-DMESH2D_BIN=/home/runner/work/elmerfem/elmerfem/build/meshgen2d/src/Mesh2D" "-DTEST_SOURCE=/home/runner/work/elmerfem/elmerfem/fem/tests/CurvedBoundaryP" "-DPROJECT_SOURCE_DIR=/home/runner/work/elmerfem/elmerfem/fem/tests" "-DSHLEXT=.so" "-DBINARY_DIR=/home/runner/work/elmerfem/elmerfem/build" "-DCMAKE_Fortran_COMPILER=/usr/bin/f95" "-DMPIEXEC=/usr/bin/mpiexec" "-DMPIEXEC_NUMPROC_FLAG=-n" "-DMPIEXEC_PREFLAGS=--allow-run-as-root" "-DMPIEXEC_POSTFLAGS=" "-DWITH_MPI=ON" "-DMPIEXEC_NTASKS=1" "-P" "/home/runner/work/elmerfem/elmerfem/fem/tests/CurvedBoundaryP/runtest.cmake"
  Directory: /home/runner/work/elmerfem/elmerfem/build/fem/tests/CurvedBoundaryP
  "CurvedBoundaryP" start time: Jul 13 21:44 UTC
  Output:
  ----------------------------------------------------------
  -- BINARY_DIR = /home/runner/work/elmerfem/elmerfem/build
  -- Extra library directories 
  CMake Error at /home/runner/work/elmerfem/elmerfem/cmake/Modules/test_macros.cmake:157 (FILE):
    FILE failed to open for reading (No such file or directory):
  
      /home/runner/work/elmerfem/elmerfem/build/fem/tests/CurvedBoundaryP/TEST.PASSED
    /home/runner/work/elmerfem/elmerfem/fem/tests/CurvedBoundaryP/runtest.cmake:3 (RUN_ELMER_TEST)
  
  
  <end of output>
  Test time =   2.55 sec
  ----------------------------------------------------------
  Test Failed.
  "CurvedBoundaryP" end time: Jul 13 21:44 UTC
  "CurvedBoundaryP" time elapsed: 00:00:02
  ----------------------------------------------------------
```

That doesn't give much to follow up on (at least for me).

I was just thinking/hoping that something would output a bit more than that to somewhere. But admittedly, I don't know if that is actually the case.

mmuetzel:
The link error with OpenMP using LLVM Clang on Ubuntu is fixed by configuring with `-DOpenMP_C_FLAGS=-fopenmp=libgomp -DOpenMP_CXX_FLAGS=-fopenmp=libgomp`.

mmuetzel:
The same set of tests fail when building with Clang independent on whether OpenMP is used.

raback:
I am looking at these great CI builds (ubuntu-clang & windows-mingw) and wondering whether they might deserve their own .yaml file. For me it seems that the status of a build is either success or failure. So if we would like to show one status flag for each build they need to have different .yaml file? Like for build.yaml we have the status in "https://github.com/ElmerCSC/elmerfem/actions/workflows/build.yaml/badge.svg". Now build.yaml gets bloated if we add all platforms there. What is your take on this? So we could have build-ubuntu-clang.yaml, build-windows-mingw.yaml etc. Just my uninformed thoughts...

mmuetzel:
The workflows for GCC vs. Clang on Ubuntu share almost all of the build instructions. So, separating them into two yaml files might lead to a lot of copy-paste and the potential for forgetting that changes might need to be applied in workflows for both.

I'm not sure if I understood your motivation correctly: Are you hesitating to add workflows for configurations that (currently) fail because that would mean that the badge would indicate that? Splitting the workflows just because they fail only puts a blanket over a potentially underlying issue imho. The configuration still fails (whether the badge is green or red doesn't make a difference).
Does ElmerFEM only "guarantee" to work for one platform with one compiler? Is it "just coincidental" if it works on other platforms or with other compilers? Is a failure on other platforms or with other compilers considered a bug?

Anyway: Since (more or less) recently, GitHub has [reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows). Maybe, it's possible to use those to keep duplication of the build rules in the yaml files to a minimum while still splitting (at least some) elements of a build matrix to different workflows...
I haven't looked into how to do that yet though.

Splitting the workflows for different platforms into separate yaml files is probably ok. Build instructions tend to differ for different platforms anyway.

raback:
My idea was just to give some more information. If we have clang build with some persisting issues it is still nice to know with a quick inspection that a commit didn't break the gcc compilation. We could have then several badges. The same info can of course be retrieved by other means. 

I am just trying to hunt why the clang fails. For some reason it seems that it cannot obtain the meshes.

raback:
I was able to get the std out of ElmerSolver out on failure. See e.g. test 55 here. 
https://github.com/ElmerCSC/elmerfem/actions/runs/10178764044/job/28153050603
Unfortunately the mesh generation is not part of the redirected output.

mmuetzel:
I have a VM with Ubuntu 24.04 where I could try and reproduce the error locally.
Do you have any hints for what I should look out to get more information for debugging? Are there any commands that I could run manually (to check whether there is more output than what ctest captures)?

raback:
First failing test is "CurvedBoundaryP". Try with:
>cd $ELMERSRC/fem/tests/CurverBoundaryP
>ElmerGrid 14 2 circle_in_square.msh -autoclean
>ElmerSolver 

For some reason it seems that the mesh conversion from Gmsh to Elmer type fails and ElmerSolver reports that it cannot find the file. I cant see the output of ElmerGrid from the CI currently.

raback:
The test is run by the runtest.cmake in the test directory:

include(test_macros)
execute_process(COMMAND ${ELMERGRID_BIN} 14 2 circle_in_square.msh -autoclean)
RUN_ELMER_TEST()

Currently the output is redirected to stdout and stderr only in the RUN_ELMER_TEST() and these are visible in CI when setting CTEST_OUTPUT_ON_FAILURE=1. Unfortunately the problems seem to occur before this.

mmuetzel:
It looks like that command segfaults:
```
.../elmerfem/build-clang/elmergrid/src/ElmerGrid 14 2 circle_in_square.msh -autoclean

Starting program Elmergrid, compiled on Jul 25 2024
Lower dimensional boundaries will be removed
Materials and boundaries will be renumbered
Nodes that do not appear in any element will be removed
Output will be saved to file circle_in_square.

Elmergrid loading data:
-----------------------
Format chosen using the first line: $MeshFormat
Gmsh version is 4.1
Loading mesh in Gmsh format 4.1 from file circle_in_square.msh
Reading 9 entities in 0D
Reading 8 entities in 1D
Reading 2 entities in 2D
Allocating for 42 knots and 97 elements.
Leading element dimension is 2
Allocating lookup table for tags of size 9
Defined 9 0DIM entities with geometric tag range [1 9]
Defined 8 1DIM entities with geometric tag range [1 8]
Defined 2 2DIM entities with geometric tag range [1 2]
Physical tag offset for 0D is 0
Geometric tag offset for 0D is 8
Reading 42 nodes in 19 blocks.
Reading 97 elements in 19 blocks.
Moving bulk elements to boundary elements
Segmentation fault (core dumped)
```

A backtrace from that crash with gdb (in case that helps):
```
Program received signal SIGSEGV, Segmentation fault.
ElementsToBoundaryConditions (data=data@entry=0x7ffffff6bd30, bound=bound@entry=0x55613740, retainorphans=retainorphans@entry=0, info=info@entry=1)
    at .../elmerfem/elmergrid/src/egmesh.c:6983
6983	    bound[j].created = FALSE;
(gdb) bt
#0  ElementsToBoundaryConditions (data=data@entry=0x7ffffff6bd30, bound=bound@entry=0x55613740, retainorphans=retainorphans@entry=0, info=info@entry=1)
    at .../elmerfem/elmergrid/src/egmesh.c:6983
#1  0x000055555558b9d7 in LoadGmshInput (data=0x7ffffff6bd30, bound=0x55613740, prefix=<optimized out>, keeporphans=0, info=info@entry=1)
    at .../elmerfem/elmergrid/src/egconvert.c:5234
#2  0x00005555555bd4b4 in main (argc=<optimized out>, argv=<optimized out>) at .../elmerfem/elmergrid/src/fempre.c:303
(gdb) p j
$1 = 0
(gdb) p bound
$2 = (struct BoundaryType *) 0x55613740
(gdb) p bound[0]
Cannot access memory at address 0x55613740
```

Maybe, `bound` is used before it is initialized?

mmuetzel:
Strange. It looks like the upper half of the address was lost somewhere (maybe overwritten on the stack???):
```
(gdb) up 2
#2  0x00005555555bd4b4 in main (argc=<optimized out>, argv=<optimized out>) at .../elmerfem/elmergrid/src/fempre.c:303
303	    if (LoadGmshInput(&(data[nofile]),boundaries[nofile],eg.filesin[nofile],
(gdb) p nofile
$3 = 0
(gdb) p boundaries
$4 = {0x555555613740, 0x0 <repeats 11 times>}
(gdb) p boundaries[nofile]
$5 = (struct BoundaryType *) 0x555555613740
(gdb) p boundaries[nofile][0]
$6 = {created = 0, nosides = 0, maxsidenodes = 0, coordsystem = 0, echain = 0, ediscont = 0, chainsize = 0, parent = 0x0, parent2 = 0x0, material = 0x0, side = 0x0, 
  side2 = 0x0, chain = 0x0, types = 0x0, discont = 0x0, normal = 0x0, elementtypes = 0x0, topology = 0x0}
(gdb) down 1
#1  0x000055555558b9d7 in LoadGmshInput (data=0x7ffffff6bd30, bound=0x55613740, prefix=<optimized out>, keeporphans=0, info=info@entry=1)
    at .../elmerfem/elmergrid/src/egconvert.c:5234
5234		  printf("unknown node %d %d in element %d\n",k,revindx[k],i);
(gdb)  p bound
$7 = (struct BoundaryType *) 0x55613740
(gdb) p bound[0]
Cannot access memory at address 0x55613740
```

mmuetzel:
I added a breakpoint early in that function and then ran `n`, `p bound`, and `p meshdim` repeatedly until I noticed that the address of `bound` changed:
```
4972	      tagoffset[meshdim] = maxphystag[meshdim];
$1075 = (struct BoundaryType *) 0x555555613740
$1076 = 0
(gdb) 
4973	      tagoffset[meshdim-1] = phystagoffset[0];
$1077 = (struct BoundaryType *) 0x555555613740
$1078 = 0
(gdb) 
4979	      GETLONGLINE;
$1079 = (struct BoundaryType *) 0x55613740
$1080 = 0
(gdb) 
4980	      if(!strstr(longline,"$EndEntities")) {
$1081 = (struct BoundaryType *) 0x55613740
$1082 = 0
(gdb) 
```

`meshdim` is still 0 at that point. So, likely `tagoffset[meshdim-1]` accesses out of bound memory...

mmuetzel:
The change in #506 avoids that segfault for me.

Issue #471, opened 2024-06-06
error: call to undeclared function 'umfpack_di_defaults'; ISO C99 and later do not support implicit function declarations


jeffhammond:
I don't see `-Werror` in my build recipe anywhere but Clang 19 errors rather than warns on these C99 violations, breaking the build.

Have you tried to build Elmer with recent versions of Clang?  Is it expected to work?

```
[ 60%] Building C object fem/src/CMakeFiles/elmersolver.dir/umf4_f77wrapper.c.o
/home/jehammond/ELMER/elmerfem/fem/src/umf4_f77wrapper.c:127:5: error: call to undeclared function 'umfpack_di_defaults'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  127 |     UMFPACK_defaults (Control) ;
      |     ^
/home/jehammond/ELMER/elmerfem/fem/src/umf4_f77wrapper.c:101:27: note: expanded from macro 'UMFPACK_defaults'
  101 | #define UMFPACK_defaults         umfpack_di_defaults
      |                                  ^
/home/jehammond/ELMER/elmerfem/fem/src/umf4_f77wrapper.c:143:12: error: call to undeclared function 'umfpack_di_symbolic'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  143 |     (void) UMFPACK_symbolic (*m, *n, Ap, Ai, Ax, Symbolic, Control, Info) ;
      |            ^
/home/jehammond/ELMER/elmerfem/fem/src/umf4_f77wrapper.c:113:27: note: expanded from macro 'UMFPACK_symbolic'
  113 | #define UMFPACK_symbolic         umfpack_di_symbolic
      |                                  ^
/home/jehammond/ELMER/elmerfem/fem/src/umf4_f77wrapper.c:158:12: error: call to undeclared function 'umfpack_di_numeric'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  158 |     (void) UMFPACK_numeric (Ap, Ai, Ax, *Symbolic, Numeric, Control, Info);
      |            ^
/home/jehammond/ELMER/elmerfem/fem/src/umf4_f77wrapper.c:104:27: note: expanded from macro 'UMFPACK_numeric'
  104 | #define UMFPACK_numeric          umfpack_di_numeric
      |                                  ^
[...]

PR #488, opened 2024-07-13
Avoid compiler warnings about implicit function declarations.

mmuetzel:
Include header that declares functions from the UMFPACK library before using them. Add path to that header to the preprocessor flags. 
Also keep similar order of linker flags independent of whether the bundled or an external UMFPACK library is used.

This warning is elevated to an error in LLVM Clang 18 (with default compiler flags). This also errors for GCC 14.


mmuetzel:
Didn't see the open issue before submitting this PR. But I guess this would solve #471.

PR #493, opened 2024-07-15
Prefer built binaries over ones installed on the system during tests

mmuetzel:
Prefer executables and libraries that have been built by ElmerFEM over binaries of the same name that might be installed on the system. Do that by prepending the paths in the build tree to an existing PATH variable.

Include path to libparpack in PATH on Windows.

MPIEXEC might be installed in a location containing spaces in the path. E.g., the default location on Windows for it is
"C:\Program Files\Microsoft MPI\Bin\mpiexec.exe". Avoid issues with spaces in that path by surrounding it with double quotes in the respective commands.


mmuetzel:
The test suite was failing for me on Windows with a cryptic error before the proposed changes. It turned out that I have libarpack installed locally in a version that is ABI incompatible with the version that ElmerFEM bundles.
The proposed changes were needed to move that library further back in the search order.

Issue #494, opened 2024-07-16
[feature request] Improve Windows Support: Shift to MSYS2 Mingw-w64 and Enhance Distribution


Foadsf:
## Background

Currently, Elmer development emphasizes WSL2 for Windows users. However, this approach has limitations:

1. WSL2 requires Hyper-V, which is unavailable on Windows Home editions.
2. Legacy WSL, while available to Home users, is deprecated and no longer supported.
3. Many users, especially in work environments, lack the privileges to use Hyper-V or either of the WSL versions.

## Proposal

To improve accessibility and ease of use for Windows users, I propose the following changes:

1. **Shift towards MSYS2 Mingw-w64**: This environment is:
   - Fully FLOSS (Free/Libre Open Source Software)
   - Available to all Windows OS users
   - Usable without admin rights
   - Can be installed portably

2. **Add MSYS2 Mingw-w64 to GitHub Actions**: This will ensure consistent building and testing on Windows platforms.

3. **Use CPack for Windows packaging**: Package Elmer for Windows and release it directly on GitHub, replacing the [separate FTP distribution](https://www.nic.funet.fi/pub/sci/physics/elmer/bin/windows/).

4. **Maintain package manager distributions**: Actively maintain and update packages for:
   - Chocolatey
   - PortableApps

## Benefits

- Wider accessibility for Windows users
- Simplified installation process
- Consistent development and distribution workflow
- Improved portability

## Next Steps

1. Investigate the feasibility of MSYS2 Mingw-w64 integration
2. Set up GitHub Actions workflow for MSYS2 Mingw-w64
3. Develop CPack configuration for Windows packaging
4. Establish processes for maintaining Chocolatey and PortableApps packages

## Discussion

I welcome thoughts and feedback from the Elmer development team and community on this proposal.

raback:
All "quick" tests now pass on github actions.

PR #496, opened 2024-07-16
Update MSYS2 job in GitHub Actions workflow

Foadsf:
- Remove unavailable mingw-w64-x86_64-openmpi package
- Disable MPI in CMake configuration

This change allows the MSYS2 Mingw-w64 build to proceed without
errors, though it disables MPI support. Future work may be needed
to re-enable MPI using an alternative package or build method.

relevant to https://github.com/ElmerCSC/elmerfem/issues/494

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -116,7 +116,7 @@ jobs:
             mingw-w64-x86_64-cmake
             mingw-w64-x86_64-openblas
             mingw-w64-x86_64-parmetis
-            make
+            base-devel
       
       - name: Configure
         run: |
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -94,3 +94,51 @@ jobs:
           echo "::group::Log from these tests"
           [ ! -f Testing/Temporary/LastTest.log ] || cat Testing/Temporary/LastTest.log
           echo "::endgroup::"
+
+
+  windows-mingw:
+    runs-on: windows-latest
+    defaults:
+      run:
+        shell: msys2 {0}
+    
+    steps:
+      - uses: actions/checkout@v4
+      
+      - uses: msys2/setup-msys2@v2
+        with:
+          msystem: MINGW64
+          update: true
+          install: >-
+            git
+            mingw-w64-x86_64-gcc
+            mingw-w64-x86_64-gcc-fortran
+            mingw-w64-x86_64-cmake
+            mingw-w64-x86_64-openblas
+            mingw-w64-x86_64-openmpi
+            mingw-w64-x86_64-parmetis
+            make
+      
+      - name: Configure
+        run: |
+          mkdir build
+          cd build
+          cmake .. -G "MSYS Makefiles" \
+            -DCMAKE_BUILD_TYPE="Release" \
+            -DBLA_VENDOR="OpenBLAS" \
+            -DWITH_OpenMP=ON \
+            -DWITH_LUA=ON \
+            -DWITH_Zoltan=OFF \
+            -DWITH_Mumps=OFF \
+            -DCREATE_PKGCONFIG_FILE=ON \
+            -DWITH_MPI=ON
+      
+      - name: Build
+        run: |
+          cd build
+          cmake --build .
+      
+      - name: Test
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -94,3 +94,50 @@ jobs:
...
+          msystem: MINGW64
+          update: true
+          install: >-
+            git
+            mingw-w64-x86_64-gcc
+            mingw-w64-x86_64-gcc-fortran
+            mingw-w64-x86_64-cmake
+            mingw-w64-x86_64-openblas
+            mingw-w64-x86_64-parmetis
+            make
```
mmuetzel: It'd probably be a good idea to install the `base-devel` group instead of only `make`.

Foadsf: @mmuetzel thanks for the comment. Do you want to share a patch or send a PR to this specific branch? I think it should be possible to give you write access to this specific branch. But I rather core devs to do that.

mmuetzel: Why can't you make that change yourself? It shouldn't be hard.

Foadsf: @mmuetzel fair point. But I did not explain myself properly. Given your experience it would be great if we could give you write access on some of the MSYS2 / Mingw-w64 / Windows-related branches. Of course, if you are interested.

.github/workflows/build.yaml
```diff
@@ -94,3 +94,53 @@ jobs:
...
+          cd build
+          cmake .. -G "MSYS Makefiles" \
+            -DCMAKE_BUILD_TYPE="Release" \
+            -DBLA_VENDOR="OpenBLAS" \
+            -DWITH_OpenMP=ON \
+            -DWITH_LUA=ON \
+            -DWITH_Zoltan=OFF \
+            -DWITH_Mumps=OFF \
+            -DCREATE_PKGCONFIG_FILE=ON \
+            -DWITH_MPI=ON
```
mmuetzel: I accidentally commented on the commit instead of on the PR. So, repeating here. Sorry for the noise.

``` suggestion
            -DWITH_MPI=ON \
            -DMPIEXEC_EXECUTABLE="C:/Program Files/Microsoft MPI/Bin/mpiexec.exe" \
```

But for this to actually pass in CI, you'd need a new step to install MSMPI.
To install it locally, you can get the installer from here: https://www.microsoft.com/en-us/download/details.aspx?id=105289 (The MS server seems to be down currently.)

I haven't looked into how that could be installed in CI though...

PS: In any case, you are missing a line continuation here.

Foadsf:
I hard reset the latest two commits on the `feat/windows-msys2-build` branch, moving them to `fix/fortran-boz-compatibility`. Sorry for the messy try and error folks.

mmuetzel:
> I hard reset the latest two commits on the `feat/windows-msys2-build` branch, moving them to `fix/fortran-boz-compatibility`. Sorry for the messy try and error folks.

If you wait for another day or two, you'll no longer need these Fortran flags. MSYS2 should have updated the `mingw-w32-*-msmpi` packages by then. The update is already in their staging repository:
https://packages.msys2.org/queue

mmuetzel:
Do you know why this PR triggers the action four times each time you push a change?

msedge_EUO9i9jKkz


One is probably because it triggers on a push on your branch, the other one is for the pull request (i.e., for a merge to the base branch).
But why are there two more?

**P.S. by Foad**  follow-up discussions [here](https://discord.com/channels/412182089279209474/697121430395027591/1262769862648791173) on Discord.

raback:
Great developmemts! I wonder if this would deserve its own .yaml file? See the comment here: https://github.com/ElmerCSC/elmerfem/pull/487#issuecomment-2260174384

raback:
The content of this PR has been manually copied to here:
https://github.com/ElmerCSC/elmerfem/blob/devel/.github/workflows/build-windows-mingw.yaml
Now each platform has its own .yaml file and they may be triggered independently. 

The build action can be seen here:
https://github.com/ElmerCSC/elmerfem/actions/workflows/build-windows-mingw.yaml

raback:
Closing this one as the msys2 build is now successfully in the CI.

PR #506, opened 2024-07-31
ElmerGrid: Avoid out of bound array index when converting from gmsh.

mmuetzel:
See also #487.


raback:
Makes sense! Maybe it should be initialized differently but good enough for now... Thanx for the fix!

PR #512, opened 2024-08-02
Add CI rules for macOS.

mmuetzel:
Use packages from Homebrew as dependencies.

The GitHub-hosted runners with macOS 13 are on Intel hardware. The ones with macOS 14 are on Apple silicon. Add runners for both (with otherwise the same configuration).

Quite a few tests are failing. I haven't looked into the reason why. (And I don't have physical access to a Mac.)

### DIFF
--- a/.github/workflows/build-macos-homebrew.yaml
+++ b/.github/workflows/build-macos-homebrew.yaml
@@ -1,0 +1,99 @@
+name: build-macos-homebrew
+on:
+  workflow_dispatch:
+  push:
+  pull_request:
+
+concurrency: ci-macos-homebrew-${{ github.ref }}
+
+jobs:
+
+  macos-homebrew:
+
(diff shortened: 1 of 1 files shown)


raback:
Great! I think we can gradually debug these various platforms this way too. Just takes some time to narrow down the issues.

raback:
Looking at the 9 failing tests in MacOS all seem to have complex valued linear systems.

raback:
My colleague had a faint recollection that these failing tests where related to the BLAS routines of MacOS. Using OpenBLAS fixed the issue according to his recollection.

mmuetzel:
> My colleague had a faint recollection that these failing tests where related to the BLAS routines of MacOS. Using OpenBLAS fixed the issue according to his recollection.

Let's give that a try: #515

PR #513, opened 2024-08-04
Remove `CMakeParseArguments.cmake` and dependent modules. 

mmuetzel:
`cmake_parse_argument` is implemented natively in CMake since version 3.5. Only previously, it was defined in a module.

ElmerFEM already requires CMake 3.10 or newer.

Remove that module and its dependencies because it is causing issues with VTK 9.1:
```
CMake Error at /usr/lib/x86_64-linux-gnu/cmake/vtk-9.1/vtkDetectLibraryType.cmake:23 (message):
  The `PATH` argument is required.
  /usr/lib/x86_64-linux-gnu/cmake/vtk-9.1/FindGLEW.cmake:19 (vtk_detect_library_type)
  /usr/lib/x86_64-linux-gnu/cmake/vtk-9.1/VTK-vtk-module-find-packages.cmake:256 (find_package)
  /usr/lib/x86_64-linux-gnu/cmake/vtk-9.1/vtk-config.cmake:150 (include)
  ElmerGUI/CMakeLists.txt:142 (FIND_PACKAGE)
```

Also, use default syntax to include the module `FindPackageHandleStandardArgs` now that there is no longer a conflict between the module that is installed with CMake and the (old) copy that was bundled with ElmerFEM.


mmuetzel:
It looks like this was reported previously by a different user on the VTK discourse forum:
https://discourse.vtk.org/t/issues-in-configuring-software-with-vtk-dependency-fails-in-vtk-cmake-files/11805

PR #507, opened 2024-08-01
CI rules for MinGW with packages and tools from MSYS2

mmuetzel:
This overhauls the CI rules for MinGW a bit to make it easier to potentially expand the build matrix in the future.

It also tries to make it more comparable with the CI rules for Ubuntu (e.g., by using the same names for the respective build steps).

I tried to have it build using the [CLANG64 environment](https://www.msys2.org/docs/environments/) of MSYS2. But that currently fails because LLVM Flang hasn't support for some feature (yet?). Maybe, we can try again after LLVM Flang 19 has been released and MSYS2 started distributing it (maybe, some time this fall).

I also tried to build with MSMPI. But a couple of test are failing. I plan on opening a (draft) PR with that on top of this one if that is ok with you.

I'm also thinking about potentially using the UMFPACK library from MSYS2 for this workflow. Potentially by adding a build matrix with the internal and the MSYS2 UMFPACK. Would that be ok?

PR #514, opened 2024-08-04
Avoid out-of-bound index.

mmuetzel:
This avoids the random crash of `rotflow` that was discussed in #507 for me.

My Fortran is pretty rusty. And I'm not familiar with the code base. So, please check if this change is sensible.

I'm not sure if I'm reading or understanding the code correctly. The "out-of-bound indices" might be coming from here:
https://github.com/ElmerCSC/elmerfem/blob/746b705c7452f2583e89b3839e3d74cf6d368530/fem/src/ElementUtils.F90#L3993-L4001
Maybe, they should be dealt with differently? Potentially, reduce `np` by the degrees of freedom of the solver somehow? (Does any of that even make sense?)

### DIFF
--- a/fem/src/SolverUtils.F90
+++ b/fem/src/SolverUtils.F90
@@ -852,7 +852,7 @@ CONTAINS
 
        np = mGetElementDOFs(pIndexes,Element)
        np = MIN(np,n)
-       Ind(1:np) = NT % BoundaryReorder(pIndexes(1:np))
+       Ind(1:np) = NT % BoundaryReorder(MIN(pIndexes(1:np),SIZE(NT % BoundaryReorder)))
        
        ! TODO: See that RotateMatrix is vectorized
        CALL RotateMatrix( Lmtr, Lvec, n, dim, NDOFs, Ind, NT % BoundaryNormals, &


raback:
It seems that bubble degrees of freedom with index larger than number of nodes cause this issue. The fix would result to the last node to be always rotated. The right thing to do would rather be not to rotate the bubble dofs. Thanx for finding the issue, I will make an alternative fix. Just testing it...

juharu:
Peter's fix is good for this case, I think. I think, this still breaks down if the bubbles are in the 
global system and not eliminated away as here, unless 'Use P normals' keyword is given. This is also true 
even without bubbles in global system and with p>1. "Use P normals" should maybe be unconditional, if 
using n-t definition for a variable with p>1 dofs or with bubbles in global system.

juharu:
Well, I might have been looking at the wrong code, as currently "UpdateGlobalEquationsVec()" seems to be 
called instead of "UpdateGlobalEquations()", even if not using vectorized assembly otherwise... 
So the fix takes care of the thing in the sense that it does not reference the array out of bounds in any case, 
still inconsistent for edge/face degrees of freedom at the n-t boundary if p>1, unless "Use P Normals=True".

PR #518, opened 2024-08-06
Run more tests regularly in CI.

mmuetzel:
The nix workflow takes approx. 30 minutes anyway. So, there is probably little harm in running more tests without extending the overall duration of a CI run by too much.


mmuetzel:
Of the six failing tests on Ubuntu, five are from ElmerIce.

I recall that you wrote that some ElmerIce tests are not supposed to pass. (Not sure if I understood that correctly.)
Are any of those tests some that are not supposed to pass anyway?

PR #527, opened 2024-08-15
Build with ParaView in CI on Ubuntu.

mmuetzel:
This enables ParaView in the CI on Ubuntu.

The `ubuntu-latest` label is still pointing to Ubuntu 22.04. Unfortunately, there is a package conflict that prevents installing ParaView on the runner:
```
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 python3-paraview : Conflicts: python3-vtk9 but 9.1.0+really9.1.0+dfsg2-3build1 is to be installed
E: Error, pkgProblemResolver::Resolve generated breaks, this may be caused by held packages.
Error: Process completed with exit code 100.
```

That package conflict is resolved in Ubuntu 24.04. Runner images with that version of Ubuntu are already hosted as a beta by GitHub: https://github.com/actions/runner-images/issues/9848

For the time being, pin the runner image to `ubuntu-24.04`. That part of the change should be reverted some time after the `ubuntu-latest` label starts pointing to `ubuntu-24.04` (so the CI won't be stuck at an eventually old version of Ubuntu).

### DIFF
--- a/.github/workflows/build-macos-homebrew.yaml
+++ b/.github/workflows/build-macos-homebrew.yaml
@@ -61,6 +61,7 @@ jobs:
             -DWITH_Zoltan=OFF \
             -DWITH_Mumps=OFF \
             -DWITH_ELMERGUI=ON \
+            -DWITH_PARAVIEW=ON \
             -DQWT_INCLUDE_DIR="${HOMEBREW_PREFIX}/opt/qwt-qt5/lib/qwt.framework/Headers" \
             -DCREATE_PKGCONFIG_FILE=ON \
             ..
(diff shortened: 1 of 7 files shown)


raback:
I wonder whether any paraview library is needed here. To my recollection "WITH_PARAVIEW" only means that you can launch paraview directly from ElmerGUI (menu & button) but this is just a system command. To be honest, the "WITH_PARAVIEW" could be there always since almost everybody uses that.

mmuetzel:
Good point. I was confused by the following:
https://github.com/ElmerCSC/elmerfem/blob/a6ad3b141e1c4a0eea43d63ccdd9f636dd0682a7/ElmerGUI/Application/src/mainwindow.cpp#L7275-L7279

But afaict, ElmerGUI isn't actually linked against ParaView. It just starts it as a separate process.

I force pushed a change that enables ParaView everywhere but doesn't install the program itself. IIUC, it isn't actually needed that ParaView is installed to be able to build the respective parts of the code.

PR #531, opened 2024-08-20
cmake: add functionality to compile only matching solver modules

juhanikataja:


juhanikataja:
At minimum one might want `ResultOutputSolve;SaveData` in `INCLUDE_ELMER_MODULES`

PR #536, opened 2024-08-26
Make work with CMake MUMPS package

scivision:
the Git commits are independent. The first is to allow expected behavior with find_package by setting the maximum policy version to 3.12. This does not affect CMake < 3.12.

The final commit adds logic to work with MUMPS CMake package https://github.com/scivision/mumps  that has become a popular way over the past several years to use MUMPS regardless of operating system or compiler or MUMPS options.

This also enables a possible future PR where MUMPS would automatically be downloaded and built if not found.


CMakeLists.txt
```diff
@@ -507,9 +498,21 @@ SET(CMAKE_REQUIRED_LIBRARIES
...
+      set(Mumps_INCLUDE_DIR ${MUMPS_INCLUDE_DIRS})
+      if(MUMPS_METIS_FOUND)
+        set(Metis_FOUND true)
+        set(Metis_LIBRARIES ${METIS_LIBRARIES})
+        set(Metis_INCLUDE_DIR ${METIS_INCLUDE_DIRS})
+      endif()
+    else()
+      FIND_PACKAGE(Mumps REQUIRED)
+    endif()
+  endif()
```
mmuetzel, line 515: To make this more similar to what is done for a similar situation with UMFPACK or CHOLMOD, would it be possible to move this to [`FindMumps.cmake`](https://github.com/ElmerCSC/elmerfem/blob/devel/cmake/Modules/FindMumps.cmake)?
See, e.g., https://github.com/ElmerCSC/elmerfem/blob/0efbcf32fa659f5d8d45b45c6b6f169700ef51ec/cmake/Modules/FindUMFPACK.cmake#L21-L26.

You'll probably need to move that upper-case to lower-case translation to that file, too.

mmuetzel:
Thank you for your contribution. 
Apart from a small review request, this looks good to me. 👍

CMakeLists.txt
mmuetzel: As background for why moving that part to `FindMumps.cmake` would be nice:
Currently, ElmerFEM doesn't install .cmake files. (It only optionally installs pkg-config files.) But if at some point in the future a feature that installs cmake modules alongside ElmerFEM should be integrated, that hunk of logic would need to be repeated wherever those installed .cmake modules would be used in downstream code.
It would be a lot cleaner if that hunk of logic would be integrated in the existing `FindMUMPS.cmake` file.

mmuetzel (commented):
Why is the commit "require CMake >= 3.12 so find_package works as expected" needed?
What doesn't work as expected with previous versions?

mmuetzel:
This PR no longer applies on a current head of the repository.

See #847 for an alternative change that better matches what is already done for similar dependencies.

PR #523, opened 2024-08-15
Avoid using MPI_IN_PLACE if it isn't supported.

mmuetzel:
MSMPI is set up such that some variables that should be shared between the MPI library and the program that is using it are dllimported via variables in common blocks. While that is likely working with DEC or Intel Fortran compilers on Windows, it doesn't work with GFortran.
See: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=47030

Afaict, that doesn't only affect MSMPI but all MPI implementations on Windows that are derived from MPICH (that also includes Intel MPI).

Add a test during configuration that checks whether the combination of Fortran compiler and MPI implementation supports the usage of `MPI_IN_PLACE`.
Define a preprocessor variable if it isn't supported.
Use a local buffer instead of `MPI_IN_PLACE` in that case.

If the combination of compiler and MPI implementation isn't affected by that issue, the changes in this PR should make no difference to the current behavior.

Fixes #521.

No additional tests are failing when applying these changes on top of #522.

PR #540, opened 2024-08-27
Fix error in configuration test for MPI_IN_PLACE.

mmuetzel:
Almost all MPI functions in Fortran need an additional argument for returning an error code. Add that argument when calling `MPI_AllReduce` in the test program.

### DIFF
--- a/cmake/Modules/testMPIcapabilities.cmake
+++ b/cmake/Modules/testMPIcapabilities.cmake
@@ -5,13 +5,12 @@
 
 if(CMAKE_CROSSCOMPILING)
   # assume it is working
-  message(STATUS "Checking whether MPI_IN_PLACE is supported with ${CMAKE_Fortran_COMPILER} -- assuming yes")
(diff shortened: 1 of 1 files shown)


mmuetzel:
Hmmm... For some reason the test now *passes* on MinGW (where it was designed to *fail*).
Somethings not quite right...

I'll try to look into that in the coming days...

mmuetzel:
It turns out `mpif90` isn't actually used to build the test file. Instead, compilation fails and no error code is set because the binary was never executed.
That might have been an issue since that test was added. (That's my bad. Sorry for that.)

mmuetzel:
I pushed another commit that links to the imported target `MPI::MPI_Fortran`. At least on Windows, that made the program build for me again.

@mk3z: Could you please check if that works correctly on Nix now? You reported issues with the changes in #523. (I erroneously assumed that those issues were fixed by setting `CMAKE_Fortran_COMPILER`. But it turns out that doesn't have any effect - and compilation of that test just failed.)

mk3z:
Seems to work on Nix

PR #544, opened 2024-08-28
Simplify build rules by linking against MPI::MPI_Fortran target

mmuetzel:
Create dummy import target `MPI::MPI_Fortran` when building without MPI. 
Link CMake targets to `MPI::MPI_Fortran` instead of using `MPI_Fortran_INCLUDE_PATH`, `MPI_Fortran_LIBRARIES`, or `MPI_Fortran_COMPILE_FLAGS`.


mmuetzel:
Found that some code that is only built with Trilinos uses MPI functions in sources written in C++.
Let's keep this as a draft until CI is set up that builds with Trilinos.

mmuetzel:
I amended the commit to also link to the C or CXX targets if the respective features are built.

I'd like to rebase this on top of #548 after that is merged to check if this is working as expected.

mmuetzel:
This seems to be working also if MPI is used from C++ sources. (And I guess it would also work if we had coverage for MPI that is used from C sources.)

Marking as ready for review.

mmuetzel:
A couple of tests are failing intermittently. They pass when they are re-run.
From the logs, it looks like that might be caused by the disc write cache not having caught up when a newly generated file should be read.
Is there a way to pause execution for a few hundred milliseconds (or whatever duration might be reasonable) in Fortran?

mmuetzel:
It might make sense to rebase this again on top of #551 (when that is merged) to make sure this works as expected with disabled MPI.

mmuetzel:
@raback: Thank you for merging #551.
It looks like this is working correctly with disabled MPI.

PR #549, opened 2024-08-29
Use MPI Fortran module in ElmerSolver if possible.

mmuetzel:
Define preprocessor macros that indicate if MPI Fortran modules are available with the detected MPI implementation.
If available prefer the Fortran module over the Fortran header. That improves compiler diagnostics and reduces the number of compiler warnings about supposedly mismatched types.

Also, fix errors that were identified by building with the MPI module.

If this change is acceptable something similar could probably also be done for ElmerIce (or parpack).


mmuetzel:
This change greatly reduces the number of (bogus) warnings from the Fortran compiler. That makes it more tangible to sift through the remaining compiler warnings and check for useful ones.

It shouldn't have any effect on functionality.

Could the changes from this PR be applied?

juharu:
This is ok, although maybe there would be some module which would cover 
most (all?) mpi usage such that the "use mpi" could be inserted to fewer (one?) 
localtion(s) ? But also fine asis by me!

juharu:
Due to this (?) commit my cmake (3.17.5) fails with this:

CMake Error at CMakeLists.txt:217 (ADD_LIBRARY):
  ADD_LIBRARY cannot create ALIAS target "Elmer::MPI_Fortran" because target
  "MPI::MPI_Fortran" is imported but not globally visible.


CMake Error at CMakeLists.txt:218 (ADD_LIBRARY):
  ADD_LIBRARY cannot create ALIAS target "Elmer::MPI_C" because target
  "MPI::MPI_C" is imported but not globally visible.


CMake Error at CMakeLists.txt:219 (ADD_LIBRARY):
  ADD_LIBRARY cannot create ALIAS target "Elmer::MPI_CXX" because target
  "MPI::MPI_CXX" is imported but not globally visible.

any idea ?

mmuetzel:
> Due to this (?) commit my cmake (3.17.5) fails with this:
> 
> CMake Error at CMakeLists.txt:217 (ADD_LIBRARY): ADD_LIBRARY cannot create ALIAS target "Elmer::MPI_Fortran" because target "MPI::MPI_Fortran" is imported but not globally visible.
> 
> CMake Error at CMakeLists.txt:218 (ADD_LIBRARY): ADD_LIBRARY cannot create ALIAS target "Elmer::MPI_C" because target "MPI::MPI_C" is imported but not globally visible.
> 
> CMake Error at CMakeLists.txt:219 (ADD_LIBRARY): ADD_LIBRARY cannot create ALIAS target "Elmer::MPI_CXX" because target "MPI::MPI_CXX" is imported but not globally visible.
> 
> any idea ?

Oof. Which version of CMake do you use? I've been testing with a pretty new version:
```
$ cmake --version
cmake version 3.30.3

CMake suite maintained and supported by Kitware (kitware.com/cmake).
```

An alternative to using ALIAS targets could be to create our own imported targets and link to the respective MPI:: targets instead.
I'll try to propose a change for that later.

In any case: Sorry for the inconvenience.

mmuetzel:
@juharu: I opened #560 that avoids aliasing an imported CMake target. Does that fix the configuration issue for you?

PR #553, opened 2024-08-30
ElmerGUI: add bounding box calculation.  Output the bounding box of the

richb2k:
nodes in the mesh to the console.  This change will harmonize ElmerGUI with ElmerGrid.  Also add the date ElmerGUI has been compiled to the console output.

### DIFF
--- a/ElmerGUI/Application/plugins/egmain.cpp
+++ b/ElmerGUI/Application/plugins/egmain.cpp
@@ -169,6 +169,8 @@ int ConvertEgTypeToMeshType(struct FemType *dat,struct BoundaryType *bound,mesh_
     return(1);
   }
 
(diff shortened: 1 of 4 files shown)


richb2k:
Attached is a small archive with three elmergrid input files for testing the bounding box output to the console.

[bbox-tests.zip](https://github.com/user-attachments/files/16918929/bbox-tests.zip)

Extract the files into a folder, open ElmerGUI, click on File, Open, and select one of the elmergrid files.  It should load and display, look through the console output for 'bounding box'.  Then click on File, Save, and save the generated elmer mesh files.  Finally, click on File, Load mesh, to load the generated mesh.  Again, look through the console output for 'bounding box'.  Repeat for the other two elmergrid input files.

PR #555, opened 2024-09-01
Build with HYPRE in CI on Ubuntu.

mmuetzel:
Install library and enable building with HYPRE.


raback:
The essential test passed fully (build.yaml). I cancelled some.

PR #557, opened 2024-09-01
Increase size of buffer to accommodate maximum number of read characters

mmuetzel:
`fgets` is set to read up to 120 characters. Increase the size of the buffer `s` to 121.


juharu:
Thanks Markus!

PR #558, opened 2024-09-02
Use ANSI C syntax for function declarations in MATC

mmuetzel:
Modern compilers emit warnings for function declarations in K&R syntax along the line of:
```
/home/runner/work/elmerfem/elmerfem/matc/src/c3d.c:115:11: warning: a function definition without a prototype is deprecated in all versions of C and is not supported in C23 [-Wdeprecated-non-prototype]
  115 | VARIABLE *c3d_gc3d(var) VARIABLE *var;
      |           ^
```

Reduce the number of compiler warnings when building MATC by switching from K&R syntax to ANSI syntax. This also allows for better compiler diagnostics during build.

There shouldn't be any change in functionality due to these changes.


juharu:
Thanks!

ealbiter:
After merging this PR, I am having problems compiling in Majaro Linux (Arch based). This is the error I am getting:
```
elmerfem/matc/src/gra_com.c: In function 'gra_gpolyline':
elmerfem/matc/src/elmer/matc.h:165:33: error: passing argument 2 of '(void (*)(int,  Point *))gra_funcs[7]' from incompatible pointer type [-Wincompatible-pointer-types]
  165 | #define MATR(ptr)    (ptr)->this->data
      |                      ~~~~~~~~~~~^~~~~~
      |                                 |
      |                                 double *
```
GCC info:
```
gcc -v                                                                                               ✔   
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-pc-linux-gnu/14.2.1/lto-wrapper
Target: x86_64-pc-linux-gnu
Configured with: /build/gcc/src/gcc/configure --enable-languages=ada,c,c++,d,fortran,go,lto,m2,objc,obj-c++,rust --enable-bootstrap --prefix=/usr --libdir=/usr/lib --libexecdir=/usr/lib --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=https://gitlab.archlinux.org/archlinux/packaging/packages/gcc/-/issues --with-build-config=bootstrap-lto --with-linker-hash-style=gnu --with-system-zlib --enable-__cxa_atexit --enable-cet=auto --enable-checking=release --enable-clocale=gnu --enable-default-pie --enable-default-ssp --enable-gnu-indirect-function --enable-gnu-unique-object --enable-libstdcxx-backtrace --enable-link-serialization=1 --enable-linker-build-id --enable-lto --enable-multilib --enable-plugin --enable-shared --enable-threads=posix --disable-libssp --disable-libstdcxx-pch --disable-werror
Thread model: posix
Supported LTO compression algorithms: zlib zstd
gcc version 14.2.1 20240805 (GCC) 
```
I think that the culprit is this commit: 53aef5e6771864f64d885e992c54c6a3b40e6cb7
@mmuetzel Let me know if you need more information

ealbiter:
Update: I made it work by adding the C flag `-Wno-error=incompatible-pointer-types` .

mmuetzel:
> Update: I made it work by adding the C flag `-Wno-error=incompatible-pointer-types` .

Yeah, that should be the work-around for the time being.
The changes from this PR enabled a few compiler diagnostics that weren't possible before. Amongst them the one you pointed out.
This particular warning was elevated to an error by default in GCC 14:
https://gcc.gnu.org/gcc-14/porting_to.html

I still haven't figured if this points to an actual issue or if a cast would be correct here...

juharu:
well, i guess it's just (mis)using the memory layouts, without regards to data types, as was 
more or less the custom at the time (in the 80's). Maybe an explicit cast can be made, or else 
copy the data to the target data type ?

mmuetzel:
Yeah. I already thought that the `data` field of that structure was used more or less like a general `void *` that could be pointing to any type. It just happens to point to a `double` most of the time...

In that case, a cast to the actual type would probably be fine to pacify the compiler diagnostic.
Alternatively, it might make sense to actually make the `data` field a `void *` type... (I haven't tested what compilers would do with that though. Maybe, it would silence the warnings. Or it might generate more warnings in other places.)

juharu:
I changed the prototypes in "~/matc" (the elmersolver version of matc) such that in doesn't complain 
'bout incompatible types when compiled with my gcc11. Shouldn't change anything functional, but 
that is moot point anyway, as the "gra"-module of matc is long since obsolite anyway... 

ElmerGrid & ElmerGUI versions of matc remain to be changed... Maybe there shouldn't be 
three separate versions ?

mmuetzel:
I'm not sure if it is the best approach to cast to a function pointer type that doesn't actually match the prototype of the function that is going to be called.

But if the `gra`-module is obsolete (and unused?) anyway, it probably isn't worth it to put much time or energy into finding the "best" solution for this...

juharu:
> gra -module is obsolete (and unused?) anyway, 
yes, definitely obsolete & unused ....

mmuetzel:
Fwiw, you'll need to downgrade that warning from an error anyway if you'd like to build Zoltan with a newer compiler. See, e.g., https://github.com/ElmerCSC/elmerfem/commit/1639975f9de691b45fdcafd38710830fccbe7d88

PR #561, opened 2024-09-04
Show more output for failing tests in full CI runs on Ubuntu.

mmuetzel:


mmuetzel:
Wait a sec. Just noticed that a potential `_np*` suffix isn't stripped correctly from the test name.
See, e.g., here: https://github.com/ElmerCSC/elmerfem/actions/runs/10686665502/job/29622448535?pr=518
```
Content of fem/tests/mgdyn_airgap_force_np2
  ---- Files ----
  Error: Process completed with exit code 2.
```

mmuetzel:
Should be better now.

PR #564, opened 2024-09-05
Split FindARPACK cmake script

ealbiter:
This PR is a follow-up of #554.

These changes allow using a mixed (p)arpack library (internal or external) configuration through the cmake variables EXTERNAL_ARPACK and EXTERNAL_PARPACK.

### Current issues
The new scripts work without problems; however, mixing libraries from different sources is not always working correctly. In my tests, I obtained the following results:
arpack source............ parpack source............. Result
Elmerfem.......................Elmerfem.....................OK
arpack-ng.......................arpack-ng...................OK
Elmerfem........................arpack-ng...................OK
arpack-ng........................Elmerfem..................Linking problems

OK: Compilation went without problems and a few failed tests.
Linking problems:
While linking libparpack,  `ld` complained about two missing symbols: `slaqrb` and `dlaqrb`. I guess that at some point, the code of arpack-ng diverged from the original source code and eliminated these functions.

I am leaving as a draft PR and would like to hear your opinions about the proposed changes.


cmake/Modules/FindPARPACK.cmake
```diff
@@ -0,0 +1,93 @@
...
+  # Fall back to manual search
+  IF(NOT PARPACK_LIBRARIES OR NOT PARPACK_INCLUDE_DIR)
+    INCLUDE(FindPackageHandleStandardArgs)
+    MESSAGE(STATUS "Manual search of parpack")
+    # Try to find PARPACK header
+    SET(ARPACKINCLUDE
+      "${ARPACK_ROOT}/include"
+      "$ENV{ARPACK_ROOT}/include"
+      "${ARPACKROOT}/include"
+      "$ENV{ARPACKROOT}/include"
```
mmuetzel: Should these variables start with "P"? E.g., `PARPACKINCLUDE` and maybe more importantly `PARPACK_ROOT` (and the others).
That would support the use-case when, e.g., a PARPACK package is installed at a different prefix than a ARPACK package.

Maybe, keep the `ARPACK*` variables as hints *after* the `PARPACK*` variables to facilitate the case when both libraries are installed at the same prefix (which is probably most common).

mmuetzel:
Looks good to me (apart from a question about a part of the new `FindPARPACK.cmake` module).

If a user selects an "incompatible" ARPACK library with the bundled PARPACK, I'd be inclined to say that it would be a user error.

Looking at the bundled versions of ARPACK and PARPACK (and potentially improving their compatibility with gfortran 10 or newer) is on my to-do list anyway. Maybe, there are options to improve "cross-version" compatibility? But even if there aren't, that shouldn't block this PR imho.

Thank you again for the contribution!

ealbiter:
> If a user selects an "incompatible" ARPACK library with the bundled PARPACK, I'd be inclined to say that it would be a user error.

I agree. So, this PR is ready to go. Thank you for your comments and observations.

mmuetzel:
Thank you for the update. Your proposed changes look good to me. 👍

juharu:
Thanks ealbiter!

cmake/Modules/FindPARPACK.cmake
```diff
@@ -0,0 +1,101 @@
...
+      "${PARPACKROOT}/lib"
+      "$ENV{PARPACKROOT}/lib64"
+      "${ARPACK_ROOT}/lib"
+      "$ENV{ARPACK_ROOT}/lib64"
+      "${ARPACKROOT}/lib"
+      "$ENV{ARPACKROOT}/lib64"
+      INTERNAL
+      )
+    FIND_LIBRARY(PARPACK_LIBRARIES NAMES parpack parpackng parpack-ng HINTS ${ARPACKLIB})
+  ENDIF(NOT PARPACK_LIBRARIES OR NOT PARPACK_INCLUDE_DIRR)
```
mmuetzel, line 73: Minor typo:
``` suggestion
  ENDIF(NOT PARPACK_LIBRARIES OR NOT PARPACK_INCLUDE_DIR)
```

Sorry for not noticing earlier.

juharu:
i already made the change -> 'devel', thanks Markus!

PR #565, opened 2024-09-06
Add QT6 support to ElmerGUIlogger and ElmerGUITester

ealbiter:
Currently, these modules do not have support for QT6. This pull request updates the corresponding CMake scripts to enable compilation using QT6.


mmuetzel:
Looks good to me. 👍

PR #567, opened 2024-09-10
Improve FindZoltan.cmake script and allow compilation with the system library

ealbiter:
The proposed changes improve the cmake script using the config script provided by the Zoltan library installed in the system. If this script is not found, a search is performed to find the required information, as in other cmake scripts in the project.
These changes will allow the use of the system library, which is impossible in the current state, as the configuration fails with this error:
```
CMake Error at fem/src/CMakeLists.txt:179 (ADD_DEPENDENCIES):
  The dependency target "/lib/libzoltan.so" of target "elmersolver" does not
  exist.
```


fem/src/CMakeLists.txt
```diff
@@ -176,7 +176,11 @@ IF(HAVE_ZOLTAN)
   IF(Zoltan_ENABLE_ParMETIS)
     TARGET_LINK_LIBRARIES(elmersolver ${ParMetis_LIBRARIES})
   ENDIF()
-  ADD_DEPENDENCIES(elmersolver ${ZOLTAN_LIBRARY})
+  IF(TARGET Zoltan::zoltan)
+    ADD_DEPENDENCIES(elmersolver Zoltan::zoltan)
```
mmuetzel: IIUC, `Zoltan::zoltan` is only a target if it is referring to an external dependency. In that case, it is always already built. Adding the build dependency here doesn't make sense if I'm not mistaken.

Please, correct me if that is wrong.

mmuetzel: Maybe, conditionally link to that target in line 175 (a couple of lines above in the same file). That might even be necessary in case the Zoltan headers are installed at a non-default prefix (that doesn't happen to be added already).

fem/src/CMakeLists.txt
```diff
@@ -176,7 +176,11 @@ IF(HAVE_ZOLTAN)
...
+  ELSE()
+    ADD_DEPENDENCIES(elmersolver zoltan)
+  ENDIF()
```
mmuetzel: What I meant with the above was the following change here:
``` suggestion
  IF(NOT TARGET Zoltan::zoltan)
    ADD_DEPENDENCIES(elmersolver zoltan)
  ENDIF()
```

And something like the following a couple of lines above:
```
  IF(TARGET Zoltan::zoltan)
    TARGET_LINK_LIBRARIES(elmersolver Zoltan::zoltan)
  ELSE()
    TARGET_LINK_LIBRARIES(elmersolver ${ZOLTAN_LIBRARY})
  ENDIF()
```

mmuetzel: Or maybe, use `USE_SYSTEM_ZOLTAN` (instead of `TARGET Zoltan::zoltan`) as the conditions to make it easier to understand the motivation for these conditions in the future...

ealbiter: Thank you for your comments. I am working on your suggestions and will update the code shortly.
Cheers

fem/src/CMakeLists.txt
```diff
@@ -172,11 +172,15 @@ IF(HAVE_PARMMG)
 ENDIF()
   
 IF(HAVE_ZOLTAN)
-  TARGET_LINK_LIBRARIES(elmersolver ${ZOLTAN_LIBRARY})
   IF(Zoltan_ENABLE_ParMETIS)
     TARGET_LINK_LIBRARIES(elmersolver ${ParMetis_LIBRARIES})
   ENDIF()
```
mmuetzel: Not sure about this. But the existing structure looks like the Zoltan library might depend on the ParMetis library.
For some linkers (e.g., `ld.bfd`) the order of linker flags matters.
If you move linking to `${ZOLTAN_LIBRARY}` down, it might be safest to keep this (conditional) block after that.

mmuetzel: Or keep this block here but move linking to the Zoltan libraries (conditionally via `${ZOLTAN_LIBRARY}` or `Zoltan::zoltan`) back up to where it was. (Sorry if my comment regarding this last week was unclear.)

CMakeLists.txt
```diff
@@ -736,6 +730,17 @@ ELSE()
...
 ENDIF()
+IF(WITH_Zoltan AND NOT USE_SYSTEM_ZOLTAN)
+  MESSAGE(STATUS " Zoltan: building bundled code")
+  MESSAGE(STATUS "------------------------------------------------")
+ELSEIF(WITH_Zoltan AND HAVE_ZOLTAN)
+  MESSAGE(STATUS " Zoltan library: " "${ZOLTAN_LIBRARY}")
+  MESSAGE(STATUS " Zoltan headers: " "${ZOLTAN_INCLUDE_DIR}")
+  MESSAGE(STATUS "------------------------------------------------")
+ELSEIF(WITH_Zoltan AND NOT ZOLTAN_FOUND)
+  MESSAGE(STATUS " Zoltan system library not found")
```
mmuetzel, line 733: Should this be a `MESSAGE(FATAL_ERROR ...)` instead of a `MESSAGE(STATUS ...)`?
If I understand correctly, we reach here if the user set to build with Zoltan (using the libraries installed on the build system) but it couldn't be found. That probably means that the user still needs to provide more configuration flags. Or something else is inconsistent...

ealbiter: From the current code version, I got the impression that the compilation could (or should?) continue even if the library was not found; that's why I left the last condition and allowed it to continue.
It may be better to stop the configuration to alert the user of the error when locating the library and correct the problem. I will make sure to update the code accordingly.

fem/src/CMakeLists.txt
ealbiter: Did you mean something like this:
```
IF(HAVE_ZOLTAN)
  IF(USE_SYSTEM_ZOLTAN)
    TARGET_LINK_LIBRARIES(elmersolver Zoltan::zoltan)
  ELSE()
    ADD_DEPENDENCIES(elmersolver zoltan)
    TARGET_LINK_LIBRARIES(elmersolver ${ZOLTAN_LIBRARY})
  ENDIF()
  IF(Zoltan_ENABLE_ParMETIS)
    TARGET_LINK_LIBRARIES(elmersolver ${ParMetis_LIBRARIES})
  ENDIF()
ENDIF()
```

mmuetzel: Yes. That looks good to me. 👍

CMakeLists.txt
mmuetzel: You are right.

IIUC, the current situation is a bit inconsistent. E.g., for `WITH_Trilinos=ON`, `WITH_MUMPS=ON` and some others, the configuration would fail if the corresponding packages wouldn't be found. On the other hand, for `WITH_Zoltan=ON`, `WITH_MMG=ON` ans some others, the configuration would succeed (with a status message hinting at the missing dependency) if those packages wouldn't be found.

But I agree with you that it is probably out of the scope for this PR to straighten these inconsistencies out.

Sorry for the noise.

mmuetzel:
Thank you!
This is looking good to me. 👍

ealbiter:
@mmuetzel  Thank you for all your comments and observations. If there are no more comments, this PR is ready.

raback:
Thanx!

PR #570, opened 2024-09-13
Optionally build with external (P)ARPACK in CI on Windows/MinGW.

mmuetzel:
PARPACK is distributed by MSYS2 now.
Use the runner that already builds against an external UMFPACK library to also test building against external ARPACK and PARPACK libraries.

### DIFF
--- a/.github/workflows/build-windows-mingw.yaml
+++ b/.github/workflows/build-windows-mingw.yaml
@@ -18,7 +18,7 @@ jobs:
 
     runs-on: windows-latest
 
-    name: MSYS2 (${{ matrix.umfpack }} UMFPACK, ${{ matrix.msystem }})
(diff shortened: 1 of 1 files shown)


mmuetzel:
A test failed on the first run on the Windows/MinGW runner but passed when it was repeated. Similarly for a different test on one of the macOS runners.

To track down the reason for issues like those, it would probably help to have (at least part of) the changes from #546 and #562. If the remaining changes are still up for debate, should I split the changes that display more information about failed tests in the logs to a new PR?

mmuetzel:
Imho, this PR would increase CI coverage. 
There already is a runner that tests building with an external UMFPACK library. I felt it most fitting to use the same runner to test with external ARPACK/PARPACK libraries.

Is this change ok?

PR #572, opened 2024-09-13
ARPACK/PARPACK: Avoid compiler error with GFortran 10 or newer. 

mmuetzel:
The `?vout` functions expect their third argument to be of rank-1. Similarly, the `p?vout` functions expect their fourth argument to be of rank-1.
Newer compilers (e.g., GFortran 10 or newer) emit an error if it is not. That error is currently downgraded to a warning by adding `-fallow-argument-mismatch` to the Fortran compiler flags for GFortran.

Avoid that error/warning in the first place by passing the respective argument with the expected rank.

Also fix two instances where `pzvout` was called with a real argument in PARPACK.

This reduces the number of compiler warnings for me by a lot.

There should be no change in functionality by this PR. (Maybe apart from the `pzvout` being erroneously used in place of `pdvout` in PARPACK.)

### DIFF
--- a/mathlibs/src/arpack/cgetv0.f
+++ b/mathlibs/src/arpack/cgetv0.f
@@ -2,13 +2,13 @@ c\BeginDoc
 c
 c\Name: cgetv0
 c
-c\Description: 
+c\Description:
 c  Generate a random initial residual vector for the Arnoldi process.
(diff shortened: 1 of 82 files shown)


juharu:
Thanks again Markus!

PR #573, opened 2024-09-17
ElmerGUI: Avoid buffer overflow.

mmuetzel:
Use a buffer that is large enough to catch the full range of possible `int` type values.

It might be unlikely to reach the number of faces for which this will become an actual issue. But it avoids the following compiler warning:
```
/home/runner/work/elmerfem/elmerfem/ElmerGUI/netgen/libsrc/csg/csgeom.cpp: In member function ‘void netgen::CSGeometry::AddSurface(netgen::Surface*)’:
/home/runner/work/elmerfem/elmerfem/ElmerGUI/netgen/libsrc/csg/csgeom.cpp:454:27: warning: ‘%d’ directive writing between 1 and 11 bytes into a region of size 9 [-Wformat-overflow=]
  454 |     sprintf (name, "nnsurf%d", cntsurfs);
      |                           ^~
/home/runner/work/elmerfem/elmerfem/ElmerGUI/netgen/libsrc/csg/csgeom.cpp:454:20: note: directive argument in the range [-2147483647, 2147483647]
```

### DIFF
--- a/ElmerGUI/netgen/libsrc/csg/csgeom.cpp
+++ b/ElmerGUI/netgen/libsrc/csg/csgeom.cpp
@@ -450,7 +450,7 @@ namespace netgen
   {
     static int cntsurfs = 0;
     cntsurfs++;
-    char name[15];
+    char name[17];
     sprintf (name, "nnsurf%d", cntsurfs);
     AddSurface (name, surf);
   }


richb2k:
Looking at the latest netgen here:

[https://github.com/NGSolve/netgen/blob/master/libsrc/csg/csgeom.cpp](url)

This issue was addressed by ' char name[20];' on line 635

Any chance of updating/synchronizing the version of netgen included in ElmerGUI, to something more recent?

mmuetzel:
The link in the previous comment doesn't work for me. Trying again with a permalink:
https://github.com/NGSolve/netgen/blob/7656211b3ee0faf3099e7ddf2b0befe61df73ebc/libsrc/csg/csgeom.cpp#L631-L638
```
  void CSGeometry :: AddSurface (Surface * surf)
  {
    static int cntsurfs = 0;
    cntsurfs++;
    char name[20];
    snprintf (name, size(name), "nnsurf%d", cntsurfs);
    AddSurface (name, surf);
  }
```

raback:
I guess this could be updated. License is LGPL in both. However, even organization of the code has changed and given that cmake just hand-picks the needed routines there would probably be quite a bit of work for the update. So as for now I will accept this PR but complete overhaul might be desirable in future.

PR #575, opened 2024-09-18
Fix: Remove `status` variable in MPI_BSEND call

juhanikataja:
Looks like a typo since standard doesn't support a status variable in `MPI_BSEND`


juharu:
Juu,  anna mennä ...

PR #577, opened 2024-09-19
Fix compilation with OCC 7.8 and newer

ealbiter:
The linking step of ElmerGUI fails with OCC 7.8 and newer due to changes in the library names TKDEIGES (formerly TKIGES) and TKDESTEP (formerly TKSTEP). 
I did not find any mention of these modifications on OCC's changelog, but see this commit: Open-Cascade-SAS/OCCT@bd651bbbd9e30fc5a73d32d11e0ea1a1821afd76, where the change seems to have been merged.
The older names were left untouched to ensure compatibility with older versions.


mmuetzel:
Thank you for this change and pointing to the corresponding upstream commit.

This looks good to me. 👍

Issue #578, opened 2024-09-19
Give option to exclude compiling internal UMFPACK in cmake


juhanikataja:
Elmer internal UMFPACK is not needed on supercomputing platforms and may cause problems with picky compiler stacks.

There should be a cmake build variable e.g. `DISABLE_INTERNAL_UMFPACK` which would disable compiling UMFPACK.

Changes would probably include: 
- Change `#define HAVE_UMFPACK` in `fem/config.h.cmake` to `#cmakedefine @HAVE_UMFPACK@`
- Modify `CMakeLists.txt` in few places

mmuetzel:
I hope you don't mind me asking which issues those are. 
Do these issues still occur if you build with a current version of UMFPACK from SuiteSparse and configure with `-DEXTERNAL_UMFPACK=ON`?

mmuetzel:
The changes from #579 should allow building without UMFPACK.

juhanikataja:
It's not actually clear what is causing the incompatibility. The passage about picky compilers was weird humor that has no place in github issues, sorry about that.

The reasoning about giving an *option* to disable internal umfpack even without external is  that 
- On a supercomputer, there is no point in having a non-parallel direct solver.
- The internal UMFPACK solver in the elmer source tree adds to increased compile time, complexity and a source for weird compilation failures.
- The internal UMFPACK version is old, newer versions are faster so for production runs one should anyhow have their own UMFPACK.
- The internal UMFPACK shouldn't be completely removed since it enables Elmer to be a fully fledged FE solver without much external dependenices.

Do we assume internally in some parts of elmer the existence of UMFPACK or is it just in the solvers?

Currently there is a crude workaround at 778d9a3b4cb1a079205564fb85c67d6a10ac5a7a where this issue stemmed from:

```
In file included from .../elmerfem/umfpack/src/umfpack/umf_analyze.c:29:
In file included from .../elmerfem/umfpack/src/umfpack/include/umf_internal.h:88:
In file included from .../elmerfem/umfpack/src/umfpack/../amd/amd_internal.h:72:
In file included from /usr/include/stdlib.h:394:
In file included from /usr/include/sys/types.h:29:
/usr/include/bits/types.h:155:34: error: typedef redefinition with different types ('struct __fsid_t' vs 'struct __fsid_t')
typedef struct { int __val[2]; } __fsid_t;
                                 ^
/usr/include/bits/types.h:155:26: note: previous definition is here
[...]

PR #579, opened 2024-09-19
Make building with UMFPACK optional.

mmuetzel:
Add new CMake variable `WITH_UMFPACK` which can be used to build without using any functions from the UMFPACK library.

The default for this new variable is `TRUE` so there is no difference in the build configuration to before unless a user explicitly configures with, e.g., `-DWITH_UMFPACK=OFF`.

This probably fixes the issue described in #578.

With these changes, some tests are failing when built with `WITH_UMFPACK=OFF` with an error like the following in the logs:
```
ERROR:: GetMatrixFormat: UMFPACK solver has not been installed.
```

Skip all tests conditional on `HAVE_UMFPACK` for which the string "UMFPACK solver has not been installed." was found in the logs.

I might have missed some tests that should be skipped but weren't run for me (because they depend on other features with which I might not have built).


raback:
I was thinking whether we should modify the code such that the many tests are run even when "umfpack" is not found. After all, direct solvers should give the same results. Maybe there could be a warning if "umfpack" not found and using "mumps" instead. Maybe this is overthingking, since probably cases when "umfpack" is not compiled are rare (never happened before this...).

mmuetzel:
Yes, you are right. The UMFPACK backend was available in any case before the changes proposed here. (Either the bundled implementation or libraries installed on the system.)
If I understood #578 correctly, it could be desirable to build Elmer without UMFPACK in some situations. Imho, it would be best to also adapt the test suite in case UMFPACK should be made an optional feature for ElmerFEM.

I don't know how to implement an alternative backend for the affected solvers.
But even if that were to be implemented, the affected tests should probably be guarded with `IF(HAVE_UMFPACK OR HAVE_MUMPS)` instead of the guard on `HAVE_UMFPACK` (only) that is proposed in this PR. After all, Mumps is optional already (and probably should be kept optional).
So, adding a guard around these tests would probably be necessary in either case.

The guards that are added in this PR could still be modified in the future in case the alternative Mumps backend for these solvers should be implemented.

raback:
I took your 1st commit without the change in the tests and added a few lines of code in branch "umfpack2" where umfpack is changed to MUMPS, if available (with a Warning). I would think that either has to be present. Direct solvers are a backbone of robustness. So this would save the several hundreds tests. Even previously umfpack is switched to some parallel solvers if trying to use it in parallel so we already did that compromise. (This actually didn't compile on my VM with WITH_UMFPACK=FALSE).

mmuetzel:
Thank you for showing the necessary changes to fall back from UMFPACK to Mumps as the direct solver.

It would still be possible for a user to configure without UMFPACK *and* without Mumps at the same time. I don't know if there are edge cases where this would still be useful? Should the configuration emit a big fat warning if a user does that? 

It will probably take a day or two before I can start looking into this again in more detail. I'll mark this PR as a draft for the time being.

mmuetzel:
I took the liberty to include your changes in this PR.

Marking as ready for review again.

mmuetzel:
I tested without UMFPACK but with MUMPS and without either UMFPACK and MUMPS. Configuration and building succeeds and all tests (with the `quick` label) pass for me in these configurations.
In the latter configuration a prominent warning is shown close to the end of the configuration (and many tests that require UMFPACK or MUMPS are skipped as expected).

raback:
Thanx Markus! This testing is now perfect since also without umfpack we have good coverage.

PR #581, opened 2024-09-22
Build with rocALUTION in CI on Ubuntu (and fix some build issues)

mmuetzel:
The rocALUTION package that is distributed by Ubuntu requires a working OpenMP implementation. 
So, don't enable it for the CI configuration using Clang.

The CMake config file that is shipped with that Ubuntu package (erroneously) contains instructions to check for an installed HIP package (which isn't distributed by Ubuntu afaict). However, that package is built without HIP support (see: https://git.launchpad.net/ubuntu/+source/rocalution/tree/debian/rules?h=ubuntu/noble-proposed#n10).
Work around that issue by manually stripping that instruction from the CMake config file distributed by Ubuntu.

The implementation of the rocALUTION solver in ElmerFEM requires MPI. 
So, don't enable it for the CI configuration without MPI. Also, fail early if a user tries to configure with rocALUTION but without MPI.

Compilation failed with errors about an undeclared type `PtrType` in `rocalution.cpp`. Checking how the corresponding vector is used, the intention was probably for this to be `int` instead.
With that change, compilation succeeds.


mmuetzel:
I opened a PR for upstream rocALUTION that should fix generating their CMake config module if it is built without support for HIP:
https://github.com/ROCm/rocALUTION/pull/248

mmuetzel:
The change was accepted upstream. 🎉

I opened the following bug report for Ubuntu regarding the issue with the CMake config file of rocALUTION:
https://bugs.launchpad.net/ubuntu/+source/rocalution/+bug/2082246

(I don't know the workflow or procedures of the Ubuntu packagers though.)

PR #587, opened 2024-09-27
Overhaul CMake build rules for AMD and UMFPACK.

mmuetzel:
Avoid manually preprocessing the source files. Some compiler flags can easily be missed which can lead to compilation issues with the preprocessed sources.

Instead, use object libraries to build the same source files with different flags.

Potentially fixes the build issues described in #578.

### DIFF
--- a/umfpack/src/amd/CMakeLists.txt
+++ b/umfpack/src/amd/CMakeLists.txt
@@ -1,73 +1,30 @@
-INCLUDE_DIRECTORIES(include)
-SET(UMFPACK_AMD_ORIG_SOURCES
-	amd_aat.c
-	amd_1.c
-	amd_2.c
-	amd_dump.c
-	amd_postorder.c
-	amd_post_tree.c
-	amd_defaults.c
-	amd_order.c
-	amd_control.c
-	amd_info.c
-	amd_valid.c
-	amd_preprocess.c
-)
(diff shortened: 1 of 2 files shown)


mmuetzel:
The build error on the macOS runners doesn't look related to the changes from this PR.

Homebrew updated their HDF5 package recently: https://github.com/Homebrew/homebrew-core/commit/961f7dc5ae02b287a71226adc7354d96561f5d36

And it looks like ElmerFEM is not the only "victim" of that change: https://github.com/Homebrew/homebrew-core/pull/170959#issuecomment-2379556024

Let's hope they'll find a proper solution for this issue.

mmuetzel:
All CI runners have passed in the last run.
It might be easier for you to check whether this resolves the issue described in #578 after it has been merged. But even if it doesn't, it is dangerous to compile preprocessed sources with different preprocessor and compiler flags. (For sure, at least `-fopenmp` is missing when preprocessing the original sources. Potentially other flags are different, too.) So, some change addressing that issue is needed in any case.

Is it ok to merge this?

mmuetzel:
@juhanikataja: Does this change fix the issue you were seeing when building the bundled UMFPACK with a Cray compiler?

mmuetzel:
Is there any chance this will be merged?

PR #588, opened 2024-10-02
Added AMR functionality for StatCurrentSolver

tapegoji:
Modified the StatCurrentSolver to have AMR functionality. Copied the functions from StatElecSolver and modifed them to match the current solver.


raback:
Thanx for the PR! I had a look and would have some questions/comments before merging.

- It seems that you need not only "Hvalue" but also "HvalueF". How are they different? If you need "HValueF" do you also need "HValue"?  
- I guess the Gmsh output solver is mostly a copy-paste from the module. Was there a lot of changes? Maybe it would make sense to have it as a library feature to avoid redundancy?
- You also have indicators for HarmonicAVSolver. Are these working in any meaningful way or just work in progress? 

Certainly the features are appreciated. Just want to understand how they plug in.

raback:
Just comparing the Adaptive.F90 routine it seems that with these dirty definitions we could use the standard GmshOutputSolver. Just need to restructure a little...

    CALL ListAddLogical(Solver % Values,'File Append',.FALSE.)
    CALL ListAddString(Solver % Values, 'Output File Name', 'gmsh_bgmesh.msh')
    CALL ListAddString(Solver % Values,'Scalar Field 1','HvalueF')

raback:
I did some minor restricting in branch “GmshAdaptive” so we do not need to have the GsmhOutputSolver in two places almost the same way. 

So the idea could be to have 
Adaptive Mesh Use Gmsh = Logical True
Mesh Command = File “gsmh -3 mesh.geo -bgm  gmsh_bgmesh.msh”
Mesh Conversion Comand = File “ElmerGrid 14 2 mesh.msh -autoclean”

This is not tested at all. Just schematics how things could work. So the StatCurrent stuff could be merged but the Adaptive stuff would come from these modifications.

tapegoji:
Sounds good to me. 
The AVsolver changes can be disregarded. I did not mean to push those changes. sorry about that.

raback:
The essential residual routines have now been copy-pasted to the current devel. There was some transition to have the API be fully done by the library. Unfortunately that required use of subroutine pointers of cray style which is an extension of the standard. So the old way is still present for now. However, I changed the names such that only one set of residual routines is needed. Use of Gmsh from library is now in "GmshAdaptive" branch and will be merged once tests  have passed.

PR #589, opened 2024-10-02
Build ElmerIce with netCDF in CI on Ubuntu

mmuetzel:


mmuetzel:
The issue with the CMake config file that is installed by netCDF for Fortran is already fixed upstream:
https://github.com/Unidata/netcdf-fortran/pull/428

Until that change is included in a released version and that version finds its way into the packages that are distributed, it is probably be best to keep using the FindNETCDF.cmake module that is part of the ElmerFEM sources.

When these newer versions of netCDF for Fortran are starting to appear, it should be easy to adapt the FindNETCDF.cmake module to pick the fixed config file up if available.

Issue #590, opened 2024-10-02
MeshUtils.F90: update to calculate mesh pieces


richb2k:
This is a followup post related to this prior post from the 3D Inductor thread:

[https://www.elmerfem.org/forum/viewtopic.php?p=31254#p31254](url)

There have been cases in the forum recently where 'Calculate Mesh Pieces' doesn't properly report the number of pieces in the mesh.

To help illustrate the issue, open the attached archive that contains an elmergrid input file, a case.sif, and a short script.  The case.sif calls for the calculation of the mesh pieces and the calculation reports that only one exists.  Examining the resulting mesh files in ElmerGUI show that there are actually two cubes, not connected to each other.  Since the cubes don't share any nodes, this is a non-conforming mesh and the calculation should show two mesh pieces exist.  The case.sif then continues and runs the Heat Equation solver, where the case.vtu results show that heat can't conduct through both cubes, since the two cubes are not connected.

The attached archive can also be run where the calculation will properly report the number of pieces, by editing the elmergrid input file and commenting out the line 'Order Nodes = 0.0 0.0 -1.0'.  Including the 'Order Nodes' command changes the numbering order of the elements, such that the pieces calculation no longer works properly.  

[cubesZ.zip](https://github.com/user-attachments/files/17235946/cubesZ.zip)

![two-cubes](https://github.com/user-attachments/assets/bfae4e7c-6f30-434f-8d23-66c8c0c204c7)

richb2k:
The issue has been fixed with PR #591

PR #591, opened 2024-10-02
MeshUtils.F90: set the value of MaxIndex based on the maximum value in the MeshPiece…

richb2k:
… vector.    Calculate mesh pieces occasionally doesn't properly report the correct number of pieces in a mesh.  Issue #590  includes a sample case showing the issue.  Looking at the code, the value of MaxIndex is set based on the last element examined at the end of the DO WHILE loop, which may give a different value depending on the ordering of the elements.   This change will set MaxIndex to the maximum value in the entire MeshPieces vector.


richb2k:
See Issue #590 for an example test case.

PR #593, opened 2024-10-14
Fix tests that run ElmerSolver multiple times.

mmuetzel:
Use CMake macro EXECUTE_ELMER_SOLVER that sets up necessary environment variables before executing the ElmerSolver binary.

This fixes the test `mgdyn_steady_quad_extruded_restart` that is currently failing in CI on Windows MinGW.
The other tests would have failed only for higher numbers of parallel processes than what is available on the free GitHub hosted runners.

### DIFF
--- a/fem/tests/InternalPartitioning/runtest.cmake
+++ b/fem/tests/InternalPartitioning/runtest.cmake
@@ -1,4 +1,4 @@
 include(test_macros)
 execute_process(COMMAND ${ELMERGRID_BIN} 1 2 angle)
-execute_process(COMMAND ${ELMERSOLVER_BIN} part.sif)
+EXECUTE_ELMER_SOLVER(part.sif)
 RUN_ELMER_TEST()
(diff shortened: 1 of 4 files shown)


raback:
Great fix!

mmuetzel:
Thank you for accepting the change.

That still leaves one test (`FrictionHeatMasked`) that is failing consistently on the Windows MinGW runner. The logs contain the following just before the error:

```
  ComputeChange: SS (ITER=1) (NRM,RELC): ( 0.93676408E-01  2.0000000     ) :: stresssolver
   DummySolver
   ***********************************
  ComputeChange: SS (ITER=1) (NRM,RELC): (       Infinity            NaN ) :: dummyroutine
  ComputeWeight: All Done
  UpdatePartitionWeight: All Done
  ForceToStress: Starting assembly...
  SetZeroAtPeriodicNodes: All Done
  ForceToStress: Assembly done
  ForceToStress: Set boundaries done
  ERROR:: ComputeChange: Numerical Error: Norm of solution appears to be NaN
```

I can't reproduce that error locally. The test passes for me locally on Windows MinGW. `NRM` and `RELC` are 0.0 after the DummySolver here (the same on Ubuntu and on Windows). `valgrind` doesn't report anything suspicious when running that test on Ubuntu.
The strange thing is that the very similar test `FrictionHeat` is passing in the CI.

Do you have an idea what could be causing that test error in the CI? Does the dummy routine do anything different comparing `FrictionHeat` and `FrictionHeatMasked`? Does the dummy solver invoke any function from a linear algebra library that might be optimized for a specific processor architecture (like OpenBLAS) or anything different that might be hardware dependent?

PR #594, opened 2024-10-15
Allow ignoring macos package manager

viraptor:
Allow building without either homebrew or macports. This allows building in other environments, for example nixpkgs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,7 +25,7 @@ ENDIF()
 
 #shamelessly borrowed from FreeCAD project: https://github.com/FreeCAD/FreeCAD/blob/master/cMake/FreeCAD_Helpers/SetupPython.cmake
 # For building on OS X
-if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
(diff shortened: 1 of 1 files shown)


mmuetzel:
It is a bit strange that this logic is part of the build system in the first place. It might be better to leave that to the user entirely...
An option to disable that check is probably a good first step in the direction of removing it entirely at some point in the future.

The `NOT DEFINED IGNORE_PACKAGE_MANAGER` is a bit odd though. It might lead to unexpected results if users configure with, e.g., `-DIGNORE_PACKAGE_MANAGER=OFF`.
Could you add an option like the following?
```
option(IGNORE_MACOS_PACKAGE_MANAGER "Don't search for packages that are installed with Homebrew or MacPorts on macOS" OFF)
```
Or maybe easier to handle in the longer run, a positive variable:
```
option(USE_MACOS_PACKAGE_MANAGER "Search for packages that are installed with Homebrew or MacPorts on macOS" ON)
```
And check its value (instead of its existence) like this:
```
if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin" AND ${USE_MACOS_PACKAGE_MANAGER})
```

viraptor:
Replaced by https://github.com/ElmerCSC/elmerfem/pull/644

PR #600, opened 2024-10-25
Added the option to save the Gmsh background mesh in point cloud format

tapegoji:
This PR modifies adaptive.f90 to allow Elmer to save the background mesh in Gmsh's .pos (point cloud) format. The .pos format provides a compact alternative for representing background mesh data, reducing file size significantly and removing the need to send the full mesh file back to Gmsh. This enhancement improves efficiency in workflows that rely on adaptive mesh refinement with Gmsh.

Key Changes:

Implemented background mesh export in .pos format in adaptive.f90.
When using the Mesh conversion command, the background mesh is saved in the RefinedMesh(n) folder, where n = 1, 2, … represents each iteration.
This folder structure aligns with MMG's mesh generation workflow, ensuring compatibility for seamless integration.
Benefits:

Reduces background mesh file size, improving data handling and storage.
Eliminates the need to transfer the complete mesh file back to Gmsh, streamlining the workflow for adaptive refinement.
Supports iterative mesh refinement workflows compatible with MMG's generation process, allowing efficient adaptive simulation setups.

### DIFF
--- a/fem/src/Adaptive.F90
+++ b/fem/src/Adaptive.F90
@@ -116,7 +116,7 @@ CONTAINS
 
     LOGICAL :: BandwidthOptimize, Found, Coarsening, GlobalBubbles, &
         MeshNumbering, DoFinalRef
-    INTEGER :: MaxDepth, MinDepth, NLen
+    INTEGER :: MaxDepth, MinDepth, NLen, MeshDim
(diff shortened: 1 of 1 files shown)


tapegoji:
Here is a test case in this repo
https://github.com/tapegoji/scc_amr

Issue #603, opened 2024-10-30
Binary Gmsh files


petsuter:
[Gmsh version 2 legacy files](https://gmsh.info/doc/texinfo/gmsh.html#MSH-file-format-version-2-_0028Legacy_0029) can be binary or text files. For text files the `$MeshFormat version-number file-type data-size` has `file-type` as `0`. Python libraries like [meshio](https://github.com/nschloe/meshio/blob/main/src/meshio/gmsh/_gmsh22.py) default to binary and write `file-type` as `1`. (The text format can be requested with `binary=False`.)

I think ElmerGrid only supports the text variant of the Gmsh 2 file format, right? Perhaps a check/warning in [egconvert.c](https://github.com/ElmerCSC/elmerfem/blob/devel/elmergrid/src/egconvert.c#L4072) could be added? (At least I ran into confusing errors with a binary Gmsh 2 file and switching to text fixed it.)

Thanks.

richb2k:
A check for a gmsh binary input file has been added.  Please test a recent (1 Nov 2024 or newer) build and let us know if it works for you.

petsuter:
I tested it and it works, thanks!

Now with the new message it's much clearer:
```
Elmergrid loading data:
-----------------------
Format chosen using the first line: $MeshFormat
Gmsh version is 2.2
Error: Gmsh input file is in binary format! Exiting.
***** ERROR: We cannot continue with this! *****
***** Gmsh input file is in binary format!
***** ...now exiting to system!
```

Just for comparison here is the old output with the same file:

```
Elmergrid loading data:
-----------------------
Format chosen using the first line: $MeshFormat
Gmsh version is 2.2
Loading mesh in Gmsh format 2.0 from file example.msh
$MeshFormat section should end to string $EndMeshFormat:

$Nodes section should end to string $EndNodes:

Allocating for 36896 knots and 0 elements.
Allocation impossible in imatrix: 1 0 0 -1
```

PR #604, opened 2024-10-30
ElmerGrid and ElmerGUI, add check for Gmsh binary input file format a…

richb2k:
…nd exit

if binary format is detected.  Both printf and bigerror are needed to print error messages to the console in both ElmerGrid and ElmerGUI, since bigerror doesn't seem to print to console in ElmerGUI.

This is in response to Issue #603.  If gmsh binary format is detected, print a message and exit.  Works for both gmsh formats 2 and 4.


raback:
Good idea! Thanx Rich!

PR #611, opened 2024-11-30
ElmerGrid, add input checks when reading gmsh msh files. 

richb2k:
Detect currently  valid gmsh formats and if ElmerGrid doesn't support the format, print an informative message, including how to fix it, and call fatal error.  This change enables reading of gmsh format 1 msh files, which wasn't working before. Gmsh type 3 msh files are detected and reported as unsupported.  As before, binary files are detected.  This PR is an extension related to Issue #603 and PR #604, where detection and reporting of binary files was added.  This completes upgrading reading of current gmsh file formats, so the user will know positively if a format is or is not supported.  A small archive with eight test cases is included.
[gmsh-all-formats-examples.zip](https://github.com/user-attachments/files/17966844/gmsh-all-formats-examples.zip)

### DIFF
--- a/ElmerGUI/Application/plugins/egconvert.cpp
+++ b/ElmerGUI/Application/plugins/egconvert.cpp
@@ -593,7 +593,7 @@ omstart:
 	mode = 1;
       }
       else if(strstr(line,"*NODE")) {
-	if(pstr = strstr(line,"NODE OUTPUT")) {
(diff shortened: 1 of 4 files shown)


richb2k:
Synchronized ElmerGui by copying in the three source files from ElmerGrid.  This will update ElmerGui with the fix for parentheses, in addition to the update to gmsh input file checking.

raback:
Thanx Rich! Looks good.

PR #614, opened 2024-12-19
Calving 3D updates

iwheel:
Updates to the 3D calving model plus changes post merge. I think the updates to the calving model were mainly on calving solvers so should have limited affect on wider functionality. Post merge changes were more complex but again more a case of getting the calving model to work with the wider Elmer updates.


raback:
Big chunk for calving. Nice going Iain!

PR #615, opened 2024-12-20
ElmerGrid, add reading Gmsh input format 4.1 in binary.  Synchronize …

richb2k:
…with ElmerGui.


richb2k:
This PR is a further extension related to Issue #603 and PR #611.  Add code to allow reading of Gmsh format 4.1 input files in binary format.  Since there were a lot of changes to the function LoadGmshInput41, the function was reformatted with two column indentation.

raback:
Nice Rich! This saves disk space quite a bit for many. I never dared to even try. However, Xmas is approaching and nobody would be reverting here would something go wrong. So have to wait over the holidays with the merge. Merry Christmas to you!

richb2k:
Hello Peter and a Merry Christmas to you too!

No worries about waiting until next year, I don't think anyone is waiting for the changes.  From doing some testing with a large case, the elmergrid binary read of the input msh file runs about twice as fast.  There should also be a bit of a speed up of gmsh operation while writing the binary msh file.

raback:
Nice addition! The reading becomes quite complicated when dealing with both ascii and binary at the same time. But life is not always pretty. Thanx!

Issue #566, opened 2024-09-09
Crash loading STL file


Timmmm:
Using the Windows binary download from Sourceforge, compiled on Feb 7 2024, I opened this STL file:

[bar.zip](https://github.com/user-attachments/files/16936512/bar.zip)

Elmer then crashes.

By the way the UX of loading meshes is not great:

1. There's File->Open and File->Load Mesh. Apparently to load a mesh you use File->Open, not File->Load Mesh?? And for some reason File->Open opens a file and File->Load Mesh opens a directory... Very confusing.
2. Neither option lists the file formats that it supports. Fair enough for File->Load Mesh, but File->Open definitely should!
3. I initially tried to load `bar.STL` (Solidworks' default extension), and *very luckily* I noticed the message in the status bar saying "make sure the file extension is lowercase", but I very easily could have missed that, and why do I even need to do that anyway? Just lowercase it yourself if you need to! At the very least you can detect if it isn't lowercase and throw up a modal error.

Timmmm:
Possibly the same as #182 though that says it has been fixed...

mmuetzel:
Thank you for your report.

Do you still see this issue when you test with a nightly build that you could download from https://www.nic.funet.fi/pub/sci/physics/elmer/bin/windows/ ?

Timmmm:
Yep.

richb2k:
PR #618  addresses most of the above issues.  ToLower was added, so the file suffix can be either upper or lower case.  

The binary stl file from Issue #566, bar.stl, now loads and is read, such as reading the proper number of facets, but then ElmerGui crashes during the meshing operation.  The binary stl file from Issue #182, 02-cores.stl, loads and is read, and then is successfully meshed and displayed.  

In other words, both binary files now load okay, but nglib has a problem meshing bar.stl and has no problem meshing 02-core.stl.

Issue #182, opened 2019-07-31
Segmentation fault opening binary stl


Fellypao:
ElmerGUI crashes when one try to open a binary stl mesh file.

Error message from terminal:

Selected nglib for stl-format
Initializing nglib
Start meshing...
Segmentation fault

The software should not crash even if binary stl import is not supported.
A binary stl sample is attached. It was generated using FreeCAD. Gmsh is able to import it without problems.
Best regards.
[02-core.stl.tar.gz](https://github.com/ElmerCSC/elmerfem/files/3451900/02-core.stl.tar.gz)

raback:
Here we rely (perhaps on too old) external libraries. Of course you're rigtht that the code should not crash even in this case.

PR #618, opened 2025-01-07
ElmerGui, add reading of binary stl files.  add toLower of file suffix,

richb2k:
so opening of either *.stl or *.STL (lower or upper case) works.

This PR is in response to issue #566 and issue #182.


raback:
Nice to have binary support for STL as well. You cannot always control how you get the files.

richb2k:
Thanks for accepting both PRs, adding binary input for gmsh and stl files should help reduce some more errors.  

Could you take a look at the Windows builds, there hasn't been new Windows installer published to the ftp site in almost a month.

PR #619, opened 2025-01-09
ElmerGui, minor fix to account for toLower() of FDNEUT, must compare the

richb2k:
input to lower case 'fdneut'.  Needed in order to successfully load any FDNEUT input file.

### DIFF
--- a/ElmerGUI/Application/src/mainwindow.cpp
+++ b/ElmerGUI/Application/src/mainwindow.cpp
@@ -1651,7 +1651,7 @@ void MainWindow::readInputFile(QString fileName) {
       tetlibInputOk = true;
     }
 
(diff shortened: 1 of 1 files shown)


raback:
FDNEUT is associated to Fidap FEM software which was one of the early CFD software living still in early 2000's but is now dead. Perhaps the ability to read these files could be removed.

richb2k:
As preparation for removing FDNEUT, perhaps a message like 'FDNEUT is no longer supported and may be removed in the future', would make sense.  Any other input formats for ElmerGrid or ElmerGui that may be candidates for the same message?

raback:
Yes, maybe it would make sense to study all the formats at once and see which may have become obsolite. On the other hand there may be some single users that build their workflow on old formats so it is difficult to know this for sure. So maybe we let things be as they are.

PR #634, opened 2025-01-22
CI: Add runner with Ubuntu on ARM64

mmuetzel:
GitHub started hosting runners with Ubuntu on ARM64 processors for open source projects for free:
https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

Add one configuration that is using these runners to the build matrix.

According to their blog post, the arm64 runners are more efficient and potentially faster than the x86_64 runners. (But it is still in a preview phase and maybe it will take some time for them to better scale and balance the load.) If that turns out to be true, it should be easy to switch more configurations in that workflow to the arm64 runners (and maybe keep only one or two running on x86_64).

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -93,10 +93,16 @@ jobs:
           submodules: recursive
 
       - name: install dependencies
+        # OpenBLAS before version 0.3.29 shipped kernels for ?NRM2 on
+        # Neoverse N2 that was inaccurate or returned NaN.  Use the
+        # reference implementation for the Ubuntu on ARM runners to avoid
+        # that issue.
+        # Consider reverting to OpenBLAS for all runners once Ubuntu ships
+        # a new enough version.
         run: |
           sudo apt -qq update
           sudo apt install -y ${{ matrix.compiler-pkgs }} cmake gfortran \
-            libopenblas-dev \
+            $([ "${{ matrix.os }}" == "ubuntu-24.04-arm" ] && echo "libblas-dev liblapack-dev" || echo "libopenblas-dev") \
             $([ "${{ matrix.mpi }}" == "with" ] && echo "libhypre-dev") \
             $([ "${{ matrix.mpi }}" == "with" ] && echo "libopenmpi-dev libmumps-dev libparmetis-dev") \
             $([ "${{ matrix.compiler }}" == "gcc" ] && echo "libsuitesparse-dev") \
@@ -119,7 +125,7 @@ jobs:
           cmake \
             -DCMAKE_BUILD_TYPE="Release" \
             -DCMAKE_INSTALL_PREFIX="${GITHUB_WORKSPACE}/usr" \
-            -DBLA_VENDOR="OpenBLAS" \
+            $([ "${{ matrix.os }}" == "ubuntu-24.04-arm" ] && echo "-DBLA_VENDOR=Generic" || echo "-DBLA_VENDOR=OpenBLAS") \
             ${{ matrix.openmp-cmake-flags }} \
             -DWITH_LUA=ON \
             $([ "${{ matrix.mpi }}" == "with" ] && echo "-DWITH_Zoltan=ON" || echo "-DWITH_Zoltan=OFF") \
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -17,40 +17,54 @@ jobs:
   ubuntu:
     # For available GitHub-hosted runners, see:
     # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners
-    runs-on: ubuntu-24.04
+    runs-on: ${{ matrix.os }}
 
-    name: ubuntu (${{ matrix.compiler }} ${{ matrix.mpi }} MPI ${{ matrix.openmp }} OpenMP)
+    name: ${{ matrix.os }} (${{ matrix.compiler }} ${{ matrix.mpi }} MPI ${{ matrix.openmp }} OpenMP)
 
     strategy:
       # Allow other runners in the matrix to continue if some fail
       fail-fast: false
 
       matrix:
+        os: [ubuntu-24.04]
         compiler: [gcc]
         mpi: [with]
         openmp: [with]
         include:
-          - compiler: gcc
+          - os: ubuntu-24.04
+            compiler: gcc
             compiler-pkgs: "g++ gcc"
             cc: "gcc"
             cxx: "g++"
+            mpi: with
             openmp: with
             openmp-cmake-flags: "-DWITH_OpenMP=ON"
-          - compiler: gcc
+          - os: ubuntu-24.04
+            compiler: gcc
             compiler-pkgs: "g++ gcc"
             cc: "gcc"
             cxx: "g++"
             mpi: without
             openmp: with
             openmp-cmake-flags: "-DWITH_OpenMP=ON"
-          - compiler: gcc
+          - os: ubuntu-24.04
+            compiler: gcc
             compiler-pkgs: "g++ gcc"
             cc: "gcc"
             cxx: "g++"
             mpi: with
             openmp: without
             openmp-cmake-flags: "-DWITH_OpenMP=OFF"
-          - compiler: clang
+          - os: ubuntu-24.04-arm
+            compiler: gcc
+            compiler-pkgs: "g++ gcc"
+            cc: "gcc"
+            cxx: "g++"
+            mpi: with
+            openmp: without
+            openmp-cmake-flags: "-DWITH_OpenMP=OFF"
+          - os: ubuntu-24.04
+            compiler: clang
             compiler-pkgs: "clang"
             cc: "clang"
             cxx: "clang++"
@@ -98,7 +112,7 @@ jobs:
         run: |
           # work around error in the rocALUTION CMake config file distributed by Ubuntu
           if [ "${{ matrix.compiler }}" == "gcc" ] && [ "${{ matrix.mpi }}" == "with" ]; then
-            sudo sed -i '/find_dependency(HIP)/d' /usr/lib/x86_64-linux-gnu/cmake/rocalution/rocalution-config.cmake
+            sudo sed -i '/find_dependency(HIP)/d' /usr/lib/$(dpkg-architecture -qDEB_HOST_MULTIARCH)/cmake/rocalution/rocalution-config.cmake
           fi
           mkdir ${GITHUB_WORKSPACE}/build
           cd ${GITHUB_WORKSPACE}/build


mmuetzel:
Oof. More tests are failing on that runner than I would have hoped.
At least some of the failing tests (`H1BasisEvaluation`, `SD_H1BasisEvaluation`, `pointload2`) also fail on macOS (Apple Silicon). Maybe, there are some assumptions somewhere that only hold for Intel/AMD processors?

I don't have physical access to ARM64 hardware. Not sure if I could help track down any of that.
Maybe, `valgrind` or some sanitizers would be able to find something that is odd also on Intel/AMD?

juharu:
Hi 
Thanks again for your work! 
Certainly shouldn't be anything processor specific here. I can also try running with valgrind. 
Have you checked it's not a stack size problem ? 
Br, Juha

juharu:
valgrind runs are clean on my unubtu laptop ...

mmuetzel:
> Have you checked it's not a stack size problem ?

I might be wrong. But wouldn't that crash the program?
It looks like the numerics don't behave as expected. E.g., the last time steps before the test `ConstantParamFunc` failed:
```
  MAIN: 
  MAIN: -------------------------------------
  MAIN: Time: 34/50:   6.800E+00
  MAIN: -------------------------------------
  MAIN: 
  HeatSolver: Solving the energy equation for temperature
  HeatSolve: 
  HeatSolve: 
  HeatSolve: -------------------------------------
  HeatSolve:  TEMPERATURE ITERATION           1
  HeatSolve: -------------------------------------
  HeatSolve: 
  HeatSolve: Starting Assembly...
  HeatSolve: Assembly:
  HeatSolve: Assembly done
  ComputeChange: NS (ITER=1) (NRM,RELC): ( 0.16009565     0.62252546E-01 ) :: heat equation
  HeatSolve: iter:    1 Assembly: (s)    0.00    0.00
  HeatSolve: iter:    1 Solve:    (s)    0.00    0.00
  HeatSolve:  Result Norm   :   0.16009565005742657
  HeatSolve:  Relative Change :    6.2252546152800785E-002
  MAIN: 
  MAIN: -------------------------------------
  MAIN: Time: 35/50:   7.000E+00
  MAIN: -------------------------------------
  MAIN: 
  HeatSolver: Solving the energy equation for temperature
  HeatSolve: 
  HeatSolve: 
  HeatSolve: -------------------------------------
  HeatSolve:  TEMPERATURE ITERATION           1
  HeatSolve: -------------------------------------
  HeatSolve: 
  HeatSolve: Starting Assembly...
  HeatSolve: Assembly:
  HeatSolve: Assembly done
  ERROR:: IterSolve: Numerical Error: System diverged over maximum tolerance.
```

I don't know what could be causing that though...

juharu:
Right, somewhat harder to figure out then, i guess, at least with nothing to test on....

juharu:
The following tests FAILED: 


All the following tests suddenly break the linear system iterator. All use the same BiCGstab- scheme. 
Don't know if there is something wrong with it, maybe miscopiled, maybe the numerical scheme is 
at fault (it's known to have some issues, the specifics escape me atm.). The remedy might be to change the 
iterator scheme, f.ex. BiCGStab -> BiCGStabL ? 

126 - ConstantParamFunc (Failed) serial 
155 - ConvergenceControl (Failed) serial transient 
301 - HeatControlExplicit (Failed) control quick serial 
433 - OptimizeSimplexFourHeaters (Failed) control serial 
434 - OptimizeSimplexFourHeatersInt (Failed) control serial 
680 - TransientCostFourHeaters (Failed) serial 
801 - fsi_box (Failed) elasticsolve fsi serial transient 

I'll try to look at these, but mostly these might be inconsequential ... 

540 - SD_H1BasisEvaluation (292 - 
292 - H1BasisEvaluation (Failed) benchmark serial 

Slightly different result norm. Someone should have a look, whether OK anyway... 

920 - pointload2 (Failed) serial

juharu:
I switched the iterator on the ConstParamFunc, ConvergenceControl, HeatControlExplicit, TransientCostFourHeater and fsi_box tests (in devel now). 
I left out the OptimizeSimplex- tests for the time being. These result norm of these tests seem to be somewhat unstable to changes of: 
1) Iterator scheme 
2) Convergence Tolerance 
3) Preconditoning 

Br, Juha

mmuetzel:
> I switched the iterator on the ConstParamFunc, ConvergenceControl, HeatControlExplicit, TransientCostFourHeater and fsi_box tests (in devel now).

Thank you for looking into this.

I rebased this PR on top of your latest changes. Let's see if this will make a difference.

mmuetzel:
That doesn't seem to have made much difference.
Still the following test failures:
```
The following tests FAILED:
	126 - ConstantParamFunc (Failed)                        serial
	155 - ConvergenceControl (Failed)                       serial transient
	292 - H1BasisEvaluation (Failed)                        benchmark serial
	301 - HeatControlExplicit (Failed)                      control quick serial
	433 - OptimizeSimplexFourHeaters (Failed)               control serial
	434 - OptimizeSimplexFourHeatersInt (Failed)            control serial
	540 - SD_H1BasisEvaluation (Failed)                     benchmark serendipity serial
	680 - TransientCostFourHeaters (Failed)                 serial
	801 - fsi_box (Failed)                                  elasticsolve fsi serial transient
	920 - pointload2 (Failed)                               serial
```

juharu:
OK, thanks, I'll have a look at the log.

juharu:
So it seems something more complicated than the iterator failure. Can I somehow misuse the git "push" triggered "Action" 
to do debugging (without waiting for too long) ?

mmuetzel:
> Can I somehow misuse the git "push" triggered "Action" to do debugging (without waiting for too long) ?

Debugging using only GitHub actions is pretty tedious unfortunately. Unfortunately, I'm not aware of any option to log into the runners and run commands manually.
I would be very interested if someone finds out how to do that.

The next "best" thing (but still pretty bad) what I mostly resort to is:
* Fork the repository to my own user account. 
* Enable running actions on that fork.
* Optionally, disable any workflows that I'm currently not interested in. (This can be done on the "three-dot-menu" after selecting the respective workflow on the "Actions" tab.)
* Optionally, modify the "configure" step of the workflow file in that fork to disable some build options (e.g., the GUI) to reduce the build time.
* Optionally, modify the "check" step of the workflow file in that fork to reduce the number of tests that are being run.
* Optionally, modify the source files in that fork to output more intermediate results.
* Then iterate until there is some useful information.

That is still quite tedious and time consuming. It would be much easier if I knew how to stop in a debugger or anything like that in a GitHub action.

juharu:
An observation about the linear system iterator fails on ARM64 platform: 

It seems that the DNRM2() - function used by all iterative methods fails at random intervals. I don't know 
the reason. I think this is linked in from the openblas() - library here ? 

If I use my one norm function to replace that, everything starts to work. Other openblas() functions don't 
seem to have problems.

juharu:
Tests 

ConstParamFunc, ConvergenceControl, HeatControlExplicit, TransientCostFourHeater, fsi_box, OptimizeSimplexFourHeaters, 
OptimizeSimplexFourHeatersInt 

all work out of box after this change. 

This is something not seen on any other platform.

mmuetzel:
> An observation about the linear system iterator fails on ARM64 platform: It seems that the DNRM2() - function used by all iterative methods fails at random intervals. I don't know the reason. I think this is linked in from the openblas() - library here ?

Good finding!

Afaict, Ubuntu 24.04 distributes OpenBLAS version 0.3.26.

I found the following commit for upstream OpenBLAS that could be related:
https://github.com/OpenMathLib/OpenBLAS/commit/3345007d8f4559fcd65ea7166695fda3a161e7ae

I.e., they changed the implementation of DNRM2 on the Neoverse N2 due to inaccuracies. That processor is used on the GitHub runners.
IIUC, the first version with that change is OpenBLAS 0.3.29.

I'll try to switch to using the reference BLAS implementation on the ARM64 runner. Maybe, that will make a difference.

mmuetzel:
Switching to the reference BLAS implementation reduced the number of failing tests from 10 to four:
```
The following tests FAILED:
	292 - H1BasisEvaluation (Failed)                        benchmark serial
	540 - SD_H1BasisEvaluation (Failed)                     benchmark serendipity serial
	549 - SD_P2ndDerivatives (Failed)                       quick serial
	920 - pointload2 (Failed)                               serial
Error: Process completed with exit code 8.
```

That's a huge step forward. 🎉

The remaining test failures are mainly the same ones that are also failing on macOS. (Not sure if that is coincidental.)

The only other (and new) one might be because numerical results are slightly different using the reference implementation compared to an optimized implementation (like OpenBLAS). I haven't checked though...

mmuetzel:
Afaict, `SD_P2ndDerivatives` (the only new one with respect to the previously failing tests) is failing with an accuracy issue:
```
 Testing 2nd derivatives with p(8)...
(202)...PASSED (303)...PASSED (404)...PASSED (504)... dfyz:  0.85232585644089298       0.85232542765237440        4.2878851858052514E-007 >   4.2616292834281920E-007
```

juharu:
OK, that doesnt seem too bad, I'll have a look ...

juharu:
Another observation about the H1BasisEvaluation -tests: only fail on ARM64 if OpenMP activated.

juharu:
So, -O3 + OpenMP to trigger the fail on ARM64 (and probably on macos-14), even if running with 1 task, otherwise and elsewhere OK.

mmuetzel:
Good finding.
Maybe, a miscompilation by `gfortran` of some OpenMP constructs for that CPU (aarch64) and/or for that OS (macOS)? It's probably not an issue in the OpenMP implementation itself because Ubuntu uses OpenMP as implemented by GCC (`libopenmp`) while macOS uses the OpenMP implementation by LLVM (`libomp`).
It's probably also possible that there is an issue with how OpenMP is used somewhere in the sources (and it just "happens" to work correctly for some combinations of operating system and used CPU).

Do you know which OpenMP constructs are causing the test errors? It's hard to find a bug report or potential workarounds with only using these broad keywords...

Is a combination of OpenMP and `-O3` needed to reproduce the issue? Or is one of them enough?

Anyway: If I understand correctly, OpenMP is only used in parts of elmerfem. And for most use cases, it is not crucial whether OpenMP is used or not. OpenMP is deactivated by default in the build rules.
Would it be ok to just stop testing OpenMP on platforms/CPUs where it turned out that it is causing errors?


While looking through the code, I found a couple of places in `NavierStokes.F90` and `SParIterSolver.F90` where `!omp` is used instead of the (documented) `!$omp`.
I don't know if those are just typos or whether those instructions are meant to be disabled.
I'll open a separate PR with fixes for these potential typos. (Not sure if that helps with the issues here though.)
Edit: See #636

juharu:
>Is a combination of OpenMP and -O3 needed to reproduce the issue? Or is one of them enough? 
Both should be active at the same time to trigger the fail. I'd second the deactivation of OpenMP 
for testing on these platforms for the time being.

mmuetzel:
> Both should be active at the same time to trigger the fail. I'd second the deactivation of OpenMP for testing on these platforms for the time being.

I opened #637 which disables building with OpenMP on the macOS CI runners.

mmuetzel:
Switched to a different configuration (without OpenMP) for which an Ubuntu on ARM runner is added in this PR.

mmuetzel:
When building without OpenMP, the only remaining test failure for Ubuntu on ARM is `pointload2`:
```
The following tests FAILED:
Errors while running CTest
	920 - pointload2 (Failed)                               serial
Error: Process completed with exit code 8.
```

That is the same test that is the only remaining test error for macOS on Apple Silicon when building without OpenMP.
That looks like a pattern.

PR #636, opened 2025-01-26
Fix typos in `!$omp` directive

mmuetzel:
Replace some occurrences of `!omp` with `!$omp`.


mmuetzel:
It doesn't look like this made a difference for the failing tests on macOS. But it might still be a good idea to fix these typos.

mmuetzel:
Even if the changes from this PR don't seem to matter for the test suite results, it's probably a good idea to fix these typos anyway.

PR #637, opened 2025-01-27
CI: Disable building with OpenMP on Apple Silicon

mmuetzel:
Building with OpenMP causes some test failures on macOS, see https://github.com/ElmerCSC/elmerfem/pull/634#issuecomment-2613840076. The affected tests currently are: `H1BasisEvaluation` and `SD_H1BasisEvaluation`.

For the time being, disable building with OpenMP on macOS in CI.

OpenMP is needed as a transitional build dependency of the SuiteSparse package from Homebrew. So, also disable building with CHOLMOD.

The logic to build with OpenMP is left in place. So, it should be easy to re-activate building with it in case the underlying issues will be fixed.

Note that the tests with OpenMP fail for optimization level `-O3` but succeed with `-O2`.

### DIFF
--- a/.github/workflows/build-macos-homebrew.yaml
+++ b/.github/workflows/build-macos-homebrew.yaml
@@ -73,11 +73,14 @@ jobs:
       - name: configure
         env:
           LDFLAGS: ${{ matrix.openmp == 'with' && format('-L{0}/opt/libomp/lib -lomp', env.HOMEBREW_PREFIX) || '' }}
+        # The tests `SD_H1BasisEvaluation` and `SD_LinearFormsAssembly` are
+        # failing on macos-13 (Intel CPU) at optimization level `-O3`.
+        # They are passing if ElmerFEM is built with optimization level `-O2`.
         run: |
           mkdir ${GITHUB_WORKSPACE}/build
           cd ${GITHUB_WORKSPACE}/build
           cmake \
-            -DCMAKE_BUILD_TYPE="Release" \
+            -DCMAKE_BUILD_TYPE=${{ matrix.os == 'macos-13' && 'RelWithDebInfo' || 'Release' }} \
             -DCMAKE_C_COMPILER=clang \
             -DCMAKE_CXX_COMPILER=clang++ \
             -DCMAKE_Fortran_COMPILER=gfortran \
--- a/.github/workflows/build-macos-homebrew.yaml
+++ b/.github/workflows/build-macos-homebrew.yaml
@@ -18,12 +18,28 @@ jobs:
 
     runs-on: ${{ matrix.os }}
 
+    name: ${{ matrix.os }} (${{ matrix.openmp }} OpenMP)
+
     strategy:
       # Allow other runners in the matrix to continue if some fail
       fail-fast: false
 
       matrix:
-        os: [macos-14, macos-13]
+        os: [macos-14]
+        openmp: [without]
+        # Building with OpenMP causes some test failures on macOS on Apple
+        # Silicon.
+        # The affected tests currently are: H1BasisEvaluation and
+        # SD_H1BasisEvaluation.
+        # For the time being, disable building with OpenMP in CI on Apple
+        # Silicon.
+        # OpenMP is also needed as a transitional build dependency of the
+        # SuiteSparse package from Homebrew.
+        # FIXME: Consider building with OpenMP again when this (potentially
+        # upstream) issue has been fixed.
+        include:
+          - os: macos-13
+            openmp: with
 
     steps:
       - name: get CPU information
@@ -49,13 +65,14 @@ jobs:
           brew install --overwrite python@3.12 python@3.13
           brew reinstall gcc
           brew install \
-            cmake libomp openblas open-mpi suitesparse \
+            cmake openblas open-mpi \
+            ${{ matrix.openmp == 'with' && 'libomp suitesparse' || '' }} \
             qwt vtk opencascade
           echo "HOMEBREW_PREFIX=$(brew --prefix)" >> $GITHUB_ENV
 
       - name: configure
         env:
-          LDFLAGS: -L${{ env.HOMEBREW_PREFIX }}/opt/libomp/lib -lomp
+          LDFLAGS: ${{ matrix.openmp == 'with' && format('-L{0}/opt/libomp/lib -lomp', env.HOMEBREW_PREFIX) || '' }}
         run: |
           mkdir ${GITHUB_WORKSPACE}/build
           cd ${GITHUB_WORKSPACE}/build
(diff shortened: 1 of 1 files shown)


mmuetzel:
With these changes, the following test (still) fails on macOS 14 (Apple Silicon):
```
The following tests FAILED:
	833 - pointload2 (Failed)                               serial
Errors while running CTest
```
For the macOS 13 runner (Intel CPU):
```
The following tests FAILED:
Errors while running CTest
	484 - SD_H1BasisEvaluation (Failed)                     benchmark serendipity serial
	486 - SD_LinearFormsAssembly (Failed)                   benchmark serendipity serial threaded
Error: Process completed with exit code 8.
```

So, `H1BasisEvaluation` and `SD_H1BasisEvaluation` no longer fail on the runner with Apple Silicon when building without OpenMP.
`SD_H1BasisEvaluation` and `SD_LinearFormsAssembly` still fail on the Intel CPU. So, no changes with respect to the "baseline" before the changes from this PR for that runner.
`pointload2` still fails on Apple Silicon.

Afaict, the remaining test failures in this PR cannot be related to OpenMP. After all, no implementation of OpenMP is installed on the runners now at all.

mmuetzel:
Since disabling OpenMP doesn't make a difference for the failing tests on Intel CPU, I updated this PR to build without OpenMP only on Apple Silicon. That way, we retain the CI coverage for OpenMP on macOS at least for Intel CPU.

juharu:
macos-13 test failures seem to be some other issue entirely, both failing test seem to pass their internal test ok, but 
overall tests fail with this: 


-- stderr: 

CMake Error at /Users/runner/work/elmerfem/elmerfem/cmake/Modules/test_macros.cmake:165 (FILE): 

FILE failed to open for reading (No such file or directory): 

/Users/runner/work/elmerfem/elmerfem/build/fem/tests/SD_H1BasisEvaluation/TEST.PASSED 

/Users/runner/work/elmerfem/elmerfem/fem/tests/SD_H1BasisEvaluation/runtest.cmake:3 (RUN_ELMER_TEST) 

0% tests passed, 1 tests failed out of 1

mmuetzel:
Looking at the CI log, the file `TEST.PASSED` doesn't exist indeed:
```
Content of fem/tests/SD_H1BasisEvaluation
  ---- Files ----
  total 488
  drwxr-xr-x  5 runner  staff     160 Jan 27 09:30 CMakeFiles
  -rw-r--r--  1 runner  staff    1840 Jan 27 09:30 CTestTestfile.cmake
  -rw-r--r--  1 runner  staff      16 Jan 27 09:24 ELMERSOLVER_STARTINFO
  -rw-r--r--  1 runner  staff   55226 Jan 27 09:24 H1BasisEvaluation.F90
  -rwxr-xr-x  1 runner  staff  143824 Jan 27 09:40 H1BasisEvaluation.dylib
  -rw-r--r--  1 runner  staff   10358 Jan 27 09:30 Makefile
  -rw-r--r--  1 runner  staff    1481 Jan 27 09:30 cmake_install.cmake
  drwxr-x---  6 runner  staff     192 Jan 27 09:46 cube
  -rw-r--r--  1 runner  staff     449 Jan 27 09:24 cube.grd
  -rw-r--r--  1 runner  staff     658 Jan 27 09:29 h1basistest.sif
  -rw-r--r--  1 runner  staff       0 Jan 27 09:46 test-stderr_1.log
  -rw-r--r--  1 runner  staff   10178 Jan 27 09:46 test-stdout_1.log
```

Could there be a situation where the test passes, but that file is still not being created?

juharu:
I'll have a further look.

mmuetzel:
Trying to follow your changes on the `act_test` branch, it looks like switching the optimization level from `-O3` to `-O2` seems to have "fixed" the test failure.

From GCC's documentation, the additional optimization passes that are activated with `-O3` are:
https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html
> `-O3`
> Optimize yet more. -O3 turns on all optimizations specified by -O2 and also turns on the following optimization flags:
> ```
> -fgcse-after-reload
> -fipa-cp-clone
> -floop-interchange
> -floop-unroll-and-jam
> -fpeel-loops
> -fpredictive-commoning
> -fsplit-loops
> -fsplit-paths
> -ftree-loop-distribution
> -ftree-partial-pre
> -funswitch-loops
> -fvect-cost-model=dynamic
> -fversion-loops-for-strides
> ```

Would it make sense to try and find out which of these flags are causing the issue? Are you suspicious it could be one of those options specifically?

juharu:
Yes, well, I have no idea which of those, one or more, could be the culprit. Could also be some combination, 
i guess. My take would be to downgrade to -O2, also for macos-14 and ARM64, and then you could also 
re-activate OpenMP.

mmuetzel:
> Yes, well, I have no idea which of those, one or more, could be the culprit. Could also be some combination, i guess. My take would be to downgrade to -O2, also for macos-14 and ARM64, and then you could also re-activate OpenMP.

Ideally, we could disable only one (or more) optimization only in the compilation units that are causing these issues.
But I understand that tracking that down is very tedious if the issues can only be reproduced with the CI runners. So, I agree that the "next best thing" might be to work around the issue only in CI (and hope that users will make sure that they won't run into the same issue).

mmuetzel:
I left the logic for adding runners without OpenMP in place to make it easier in case we would ever want to add them at some point in the future. Currently, the build matrix that doesn't contain that configuration though.

Like you proposed, I changed the build type to "RelWithDebInfo" for the Apple Silicon and Intel CPU runners.
Let's see if that avoids these pesky test errors.

mmuetzel:
In the latest iteration (with OpenMP, with `-O2`), the following tests are failing on macOS 14 (Apple Silicon) again:
```
The following tests FAILED:
	238 - H1BasisEvaluation (Failed)                        benchmark serial
	484 - SD_H1BasisEvaluation (Failed)                     benchmark serendipity serial
	833 - pointload2 (Failed)                               serial
Error: Process completed with exit code 8.
```

Maybe, will still need to build without OpenMP on Apple Silicon and can revert to `-O3` for that runner.

The macOS 13 runner (Intel CPU) passed all tests in the test suite now. So, we could probably keep building with OpenMP and `-O2` for that one.

mmuetzel:
The only failing test is for macOS 14 (Apple Silicon):
```
The following tests FAILED:
	833 - pointload2 (Failed)                               serial
Error: Process completed with exit code 8.
```

That is the same test that also failed for the Ubuntu on ARM runner.

It didn't matter whether ElmerFEM was built with or without OpenMP or which optimization level was used. That particular test always failed on macOS 14 (Apple Silicon).

mmuetzel:
What is really strange about the failing `pointload2` test:
According to the CI log, the file `TEST.PASSED` exists after that test was run:
```
Content of fem/tests/pointload2
  ---- Files ----
  total 88
  drwxr-xr-x  4 runner  staff    128 Jan 28 07:41 CMakeFiles
  -rw-r--r--  1 runner  staff   1733 Jan 28 07:41 CTestTestfile.cmake
  -rw-r--r--  1 runner  staff      9 Jan 28 07:39 ELMERSOLVER_STARTINFO
  -rw-r--r--  1 runner  staff   7834 Jan 28 07:41 Makefile
  -rw-r--r--  1 runner  staff      2 Jan 28 07:55 TEST.PASSED
  -rw-r--r--  1 runner  staff   1669 Jan 28 07:41 case.sif
  -rw-r--r--  1 runner  staff   1461 Jan 28 07:41 cmake_install.cmake
  drwxr-x---  6 runner  staff    192 Jan 28 07:55 square
  -rw-r--r--  1 runner  staff    637 Jan 28 07:39 square.grd
  -rw-r--r--  1 runner  staff      0 Jan 28 07:55 test-stderr_1.log
  -rw-r--r--  1 runner  staff  10705 Jan 28 07:55 test-stdout_1.log
```

If I understood correctly, that file is only supposed to be created for succeeding tests. But it is still marked as failing.

Is something different for that test compared to the others?

mmuetzel:
Never mind. If the file exists, it depends on its content whether the test is counted as passing. (The file content must be `1`.)

The test seems to be failing due to an accuracy issue:
```
  WARNING:: CompareToReferenceSolution: Solver 1 FAILED:  Norm = 7.25882728E-01  RefNorm = 7.25476270E-01
  CompareToReferenceSolution: Relative Error to reference norm: 5.602639E-04
```

(In my opinion, the deviation doesn't look too bad. But tbh, I can't tell what is an acceptable deviation.)

raback:
Thanx for looking into this! I checked the pointload2 case. It is related to a point-like volumetric source that can be given to a flow equation. The issue is that it was very sensitive. I activated a "stokes" mode, decreased resolution and took up iterative solvers to fight the sensitivity. In the original case even MUMPS and Umfpack gave a little different solutions.

mmuetzel:
Thanks!

I rebased this PR on top of your change. Let's see if that helps for the M1 runner.

PR #585, opened 2024-09-26
Rename folder for tests to match the name of the tests.

mmuetzel:
For most tests, the name of the folder matches the name of the test.
This was probably just a typo.

PR #638, opened 2025-01-27
Rename some test folders to match the name of the ctest that they contain

mmuetzel:
That makes it easier to display relevant information on test failures in the CI logs.


mmuetzel:
This is similar to what was done in #585 and should be very low risk.

raback:
Thanx! This is a good policy.

PR #640, opened 2025-01-28
CI: Remove automatic triggers for act_test workflow.

mmuetzel:
The `act_test` workflow is meant for manual debugging. It probably doesn't need to run on a "regular" push or pull request.

Do not trigger automatically on push or pull request. Only keep the manual workflow dispatch trigger.


juharu:
Thanks Markus!

PR #643, opened 2025-02-06
Remove incorrect status argument from some MPI_BSEND calls

jthies:
MPI_BSEND does not have a status argument and the code did not compile when enabling the AMGX interface.


juharu (approved):
Thanks Jonas!

PR #644, opened 2025-02-06
CMake: Allow users to ignore package managers on macOS.

mmuetzel:
Addresses the issue discussed in #594.

The original contributor (@viraptor) hasn't reacted on the feedback in a few month. I hope it is ok to pick it up and propose an alternative implementation.

The currently existing rules point to FreeCAD were that part was originally "borrowed". But that project dropped that part of the build rules entirely:
https://github.com/FreeCAD/FreeCAD/pull/8400

Maybe, ElmerFEM should do the same and also drop the package manager thing on macOS. (I'd be in favor of that.)

This PR proposes a less intrusive change that allows the user to optionally skip that part (and does so in the CI rules on macOS).

Please, let me know if you prefer to remove that section of the build rules instead. (Like I wrote, imho that would probably be preferable.)


viraptor:
Sorry, missed the original comment. That looks better than my idea and should be just fine for nixpkgs builds.
Thanks for trying again ❤️

PR #645, opened 2025-02-09
CMake: Add CUDA include directories if AMGX is to be used

jthies:
Required for compiling amgx.c,
CUDA_INCLUDE_DIRS is set by find_package(CUDA) in FindAMGX.cmake


raback:
Thanx Jonas!

PR #675, opened 2025-06-13
Feature/ed circ

ettaka:
I did some review and documentation for our electrodynamic models (AV).
I think I also found a bug in MagnetoDynamics2D.

Let me know if this can be merged.


mmalinen:
Around lines 2175 it is now defined

`cmplx_val = IP % s(t)*detJ*SUM(Wbasis(j,:)*gradv)`

which is not feasible if dim == 2. If dim == 2, the value of cmplx_val should be set differently? This problem is felt when running the 2-D tests

mgdyn2D_compute_average_b
mgdyn2D_compute_bodycurrent
mgdyn2D_compute_complex_power

PR #680, opened 2025-08-08
Patch XIOS integration for YAC coupling

BenjaminRodenberg:
@tzwinger


fem/src/mpi_handshake.c
BenjaminRodenberg, line 1: Instead of adding `fem/src/mo_mpi_handshake_stub.F90` I could also add some guards here and use the environment variable `HAVE_MPI`. If there is a more fitting approch or environment variable to use, please let me know.

fem/src/SParIterComm.F90
```diff
@@ -230,6 +255,53 @@ FUNCTION ParCommInit( ) RESULT ( ParallelEnv )
...
+    ENDIF
+#endif
+
+IF (NUM_GROUPS > MAX_NUM_GROUPS) THEN
+    WRITE( Message, * ) 'Too many communication groups defined.'
+    CALL Fatal( 'ParCommInit', Message )
+ENDIF
+
+! Do comm splitting using handshake
+CALL mpi_handshake(MPI_COMM_WORLD, GROUP_NAMES(1:NUM_GROUPS), GROUP_COMMS(1:NUM_GROUPS))
```
BenjaminRodenberg, line 301: I will also simplify this part and the macros using `ELMER_COLOUR` a bit: Only use `mpi_handshake` if `HAVE_XIOS` or `HAVE_YAC`; if none of the two is used it is a better idea to keep the `MPI_COMM_SPLIT` from below for compatibility with existing software that relies on the colouring.

BenjaminRodenberg (commented):
This is just a brief review to point to currently open points and how I would like to address them.

PR #698, opened 2025-10-06
Fix CI on macOS runners

mmuetzel:
Apple and Homebrew ended support for macOS 13. 
Also, GitHub is deprecating their hosted macOS 13 runners. 

Switch those runners over to (newly) available hosted macOS 15 runners.
Use their hosted `macos-15-intel` runner as probably the last one to run on Intel x86_64 architecture.

Additionally, use the expat package from Homebrew.


raback:
Thanx for the fix! This would have been a tough one!

PR #700, opened 2025-10-07
Update MMG Free_all and Init_mesh calls to use the latest API and fix Hypre import

tmyllaeri:
I noticed that Elmer compilation fails with the latest version of both Hypre and MMG.

This PR changes the MMG calls to follow the example from MMG "develop"branch:  https://github.com/MmgTools/mmg/commits/develop/libexamples/mmg3d/adaptation_example0_fortran/example0_a/main.F90

NOTE: This change is most likely not compatible with MMG master branch which is roughly year outdated from "develop"

HYPRE compilation is fixed by simple changing a single import from "krylov.h" to "_hypre_krylov.h"

### DIFF
--- a/elmerice/Solvers/CalvingRemeshMMG.F90
+++ b/elmerice/Solvers/CalvingRemeshMMG.F90
@@ -575,9 +575,9 @@ SUBROUTINE CalvingRemeshMMG( Model, Solver, dt, Transient )
       mmgLs  = 0
       mmgMet  = 0
 
-      CALL MMG3D_Init_mesh(MMG5_ARG_start, &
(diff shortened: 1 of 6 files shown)


tmyllaeri:
Build tests are failing. At least the Hypre import fails, though locally this seems to work

```
/home/runner/work/elmerfem/elmerfem/fem/src/SolveHypre.c:48:10: fatal error: _hypre_krylov.h: No such file or directory
   48 | #include "_hypre_krylov.h"
```

raback:
We should probably make some #if's for the old vs. new MMG stuff. The release version is still using the old and if develop is using the new it is difficult to alter between the two otherwise.

PR #704, opened 2025-10-14
Yac coupling with xios

ClemensSchannwell:
This PR contains changes pertaining to the inclusion of YAC into Elmer as well as running Elmer together with YAC and  XIOS. In addition, it allows to run a detached XIOS server for multiple Elmer domains while keeping the "component name" the same for YAC. 

I am not sure why there are so many files changes. In my mind changes should be fairly minor and restricted to a handful of files...


tzwinger:
We close this one as Elmer versions seem to have diverged and hand-pick the two changed files

PR #712, opened 2025-10-21
Add MPRGP iterative method

dancaseb:


fem/src/IterativeMethods.F90
```diff
@@ -2142,29 +2141,33 @@ SUBROUTINE MPRGP(n, x, b, c, epsr, maxit, Gamma, adapt, bound, &
...
-          ! CG-like step
+          ! CG step
 
-          !CALL C_matvec(p, Ap, ipar, matvecsubr)
           CALL matvecsubr( p, Ap, ipar )        
           rtp = dotprodfun(n, z, 1, g, 1) ! residual * p
           pAp = dotprodfun(n, p, 1, Ap, 1)
 
           IF (ABS(pAp) < eps_local) THEN
-            CALL Fatal('itermethod_mprgp','p''*A*p nearly zero, stopping')
```
dancaseb, line 2153: I need to investigate this, just realized my code usually stops because of pAp being close to zero and not that the norm of gp is close to 0. It seems to give correct results though.

PR #715, opened 2025-10-28
Implementing reluctivity function in magneto dynamics 2D with a test case with an anisotropic BH reluctivity function. Squashed commits. [Feature]

KasraDjad:
This implementation only works when handle assembly is true. It uses the same same functions and subroutines that whitneyAV solver does to achieve this as well as the same logic.

I have also included a test case with an anisotropic BH reluctivity function, to prove the functionality.

Commits were condensed as requested in forum. 

KD


raback:
Thanx for the extension! Nice looking test case also.

PR #717, opened 2025-11-03
YAC-coupling: Refactor config file checks

BenjaminRodenberg:
* Define config file names through `PARAMETERS` at beginning of the file
* Check existence of config files at beginning to make this independent of other checks and features (less error prone)
* Minor changes in debugging statements

### DIFF
--- a/fem/src/SParIterComm.F90
+++ b/fem/src/SParIterComm.F90
@@ -214,9 +214,10 @@ CONTAINS
 
     INTEGER :: ierr
     INTEGER :: req, prov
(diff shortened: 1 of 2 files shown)


ClemensSchannwell:
Yes, this looks good and I saw no regression with respect to the branch `yac-coupling` in all my test cases.

BenjaminRodenberg:
@ClemensSchannwell Thanks for checking and the feedback.

I just took care of squashing everything into a single commit to keep the history a bit more tidy. (that's why the CI is currently running again)

@tzwinger Ready to merge :rocket:

PR #722, opened 2025-11-20
WIP: YAC coupling: Bugfix MPI Handshake

BenjaminRodenberg:
This PR fixes a regression we recently observed in https://github.com/ElmerCSC/elmerfem/tree/yac-coupling after merging https://github.com/ElmerCSC/elmerfem/commit/14a6ab2f0d0905a9c168d89abe25fc99839db9a0.

@ClemensSchannwell will perform some final tests in his production setup. After that we can merge this PR into `yac-coupling`

Main point: There seems to be a bug in the MPI Handshake implementation in our patched version of XIOS. A work-around for this problem is using the MPI Handshake implementation provided by YAC (see https://github.com/ElmerCSC/elmerfem/commit/8c60381da36307fec51b4ab1061ca1be5d77407f). I additionally, added a safeguard to avoid using the XIOS mpi handshake implementation (see https://github.com/ElmerCSC/elmerfem/commit/f97269433c7fb78acddbf1da4bb2e08e8c1f0612). This part needs a fix upstream.

### DIFF
--- a/fem/src/SParIterComm.F90
+++ b/fem/src/SParIterComm.F90
@@ -94,12 +94,12 @@ MODULE SParIterComm
 # endif
 #endif
 
-#ifdef HAVE_XIOS
-  USE XIOS, ONLY: xios_initialize, xios_context_finalize, xios_finalize
-#endif
-
 #ifdef HAVE_YAC
   USE elmer_coupling, ONLY: coupling_init, coupling_finalize, coupling_setup
+#endif
+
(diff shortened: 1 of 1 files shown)


ClemensSchannwell:
I have now extensively tested this bugfix and can confirm that all of the functionality is back up and running. I tested Elmer without XIOS and YAC, with only XIOS, and both XIOS and YAC. So from my side this is ready for merging!

PR #724, opened 2025-11-26
Fix test HydrostaticNSVec-ISMIP-HOM-C on Windows.

mmuetzel:
On Windows, shared libraries have the file extension .dll (compared to .so on, e.g., Linux).
Remove the hard-coded file extension from the .sif file of that test to fix a test error on Windows.


raback:
Thanx! Makes sense.

PR #726, opened 2025-11-28
Remove VTK Java wrappings

samiilvonen:
Filter out VTK Java wrappers from the list of libraries that are linked into ElmerGUI. This dependency complicates the building of Debian/Ubuntu packages and we do not use the Java wrappers.


raback:
Nice fix Sami!

PR #738, opened 2025-12-23
Fix build on Ubuntu if building with XIOS and Elmer/Ice

BenjaminRodenberg:
* Add XIOS in targets in `CMakeLists.txt` and fix order. (1)
* Only import ISO_C_BINDING where needed to avoid ambiguity. (2)
* Use INTRINSIC keyword.

(1) fixes the following compile time error:

```sh
[ 41%] Building Fortran object fem/src/CMakeFiles/elmersolver.dir/ZirkaHysteresis.F90.o
.../elmerfem/fem/src/ZirkaHysteresis.F90:970:14:

  970 |   TYPE(C_PTR) :: c_zirka_ptr
      |              1
Error: Type name ‘c_ptr’ at (1) is ambiguous
.../elmerfem/fem/src/ZirkaHysteresis.F90:978:19:

  978 |   call C_F_POINTER(c_zirka_ptr, ZirkaModelPtr)
      |                   1
Error: Argument CPTR at (1) to C_F_POINTER shall have the type TYPE(C_PTR)
```

(2) fixes the following linker error:

```sh
...
[ 56%] Linking Fortran executable ElmerSolver_mpi
[ 56%] Built target BeamSolver3D
[ 56%] Built target Acoustics
[ 56%] Built target ArteryOutlet
[ 56%] Built target CyclicConvergence
[ 56%] Built target CompressibleNS
[ 56%] Built target CoordinateTransform
/usr/bin/ld: .../lib/libxios.so: undefined reference to `nc_inq_vardimid'
...
/usr/bin/ld: .../lib/libxios.so: undefined reference to `nc_def_var_fill'
collect2: error: ld returned 1 exit status
make[2]: *** [fem/src/CMakeFiles/Solver_TGT.dir/build.make:123: fem/src/ElmerSolver_mpi] Error 1
make[1]: *** [CMakeFiles/Makefile2:21638: fem/src/CMakeFiles/Solver_TGT.dir/all] Error 2
make: *** [Makefile:166: all] Error 2
```

Note: I only ran into these problems when building with `-DWITH_XIOS=ON` on Ubuntu 24.04; when building on Levante or without XIOS I do not observe these problems.


elmerice/Solvers/CMakeLists.txt
```diff
@@ -152,16 +152,18 @@ IF(NOT(WIN32))
...
-IF(HAVE_NETCDF)
-  TARGET_LINK_LIBRARIES(ElmerIceSolvers ${NETCDF_LIBRARIES})
-ENDIF()
-IF(HAVE_HDF5)
-TARGET_LINK_LIBRARIES(ElmerIceSolvers ${PHDF5_LIBRARIES})
-ENDIF()
 IF(HAVE_XIOS)
   INCLUDE_DIRECTORIES(${XIOS_INCLUDE_DIR})
-  TARGET_LINK_LIBRARIES(ElmerIceSolvers ${XIOS_LIBRARIES})
+  TARGET_LINK_LIBRARIES(ElmerIceSolvers ${XIOS_LIBRARIES} ${NETCDF_LIBRARIES} ${PHDF5_LIBRARIES})
```
mmuetzel: Is `PHDF5_LIBRARIES` guaranteed to be set to something meaningful if `HAVE_XIOS` evaluates to true?

Would it be safer to just move the existing case `IF(HAVE_XIOS)` ahead of the `IF(HAVE_NETCDF)`? (And maybe also fix the indentation of the `IF(HAVE_HDF5)` body while you are at it.)

If XIOS actually has a hard dependency on PHDF5, maybe adjust the following condition to have a clearer error message: 
https://github.com/ElmerCSC/elmerfem/blob/443cd5b2c07e5627fbc912094a73de4e50d21b3e/CMakeLists.txt#L493-L496

fem/src/CMakeLists.txt
```diff
@@ -254,21 +254,25 @@ IF(WITH_MPI)
...
+  IF(XIOS_LIBRARIES)
+    # XIOS depends on NetCDF and HDF5, since linking order matters, add all in one go
+    LIST(APPEND ELMERSOLVER_LIBRARIES
+         ${XIOS_LIBRARIES}
+         ${NETCDF_LIBRARIES}
+         ${PHDF5_LIBRARIES})
+  ELSEIF(NETCDF_LIBRARIES)
+    LIST(APPEND ELMERSOLVER_LIBRARIES
+         ${NETCDF_LIBRARIES}
+         ${PHDF5_LIBRARIES})
```
mmuetzel: The logic in the root CMakeLists.txt file suggests that PHDF5 is an optional dependency of NetCDF:
https://github.com/ElmerCSC/elmerfem/blob/443cd5b2c07e5627fbc912094a73de4e50d21b3e/CMakeLists.txt#L399-L417

Are you sure that it is safe to append `PHDF5_LIBRARIES` unconditionally here?

If PHDF5 is indeed a hard dependency of NetCDF, the handling of it should be adapted in the root CMakeLists.txt file should be adjusted instead.

fem/src/CMakeLists.txt
```diff
@@ -254,21 +254,25 @@ IF(WITH_MPI)
...
-    LIST(APPEND ELMERSOLVER_LIBRARIES ${NETCDF_LIBRARIES})
-    IF(PHDF5_LIBRARIES)
-      LIST(APPEND ELMERSOLVER_LIBRARIES ${PHDF5_LIBRARIES})
-    ENDIF()
+  IF(XIOS_LIBRARIES)
+    # XIOS depends on NetCDF and HDF5, since linking order matters, add all in one go
+    LIST(APPEND ELMERSOLVER_LIBRARIES
+         ${XIOS_LIBRARIES}
+         ${NETCDF_LIBRARIES}
+         ${PHDF5_LIBRARIES})
```
mmuetzel: Same as above.

elmerice/Solvers/CMakeLists.txt
BenjaminRodenberg: > Is `PHDF5_LIBRARIES` guaranteed to be set to something meaningful if `HAVE_XIOS` evaluates to true?

Thanks for pointing this out. `IF(HAVE_XIOS)` only `NETCDF_LIBRARIES` is guaranteed to contain something reasonable.

I experimented a bit with building Elmer for `-DWITH_XIOS=ON` while turning PHDF5 on/off (commenting out `FIND_PACKAGE(HDF5)`) to understand better if there is a hard dependency on PHDF5 . If PHDF5 is missing this will raise the following warning:

```sh
CMake Warning at CMakeLists.txt:417 (MESSAGE):
    
   Missing: >HDF5_INCLUDE_DIR< , >HDF5_LIBRARY<,  >HDF5F_LIBRARY< 
   some functionalities will be disabled
```

Building with this configuration will also complete successfully. Meaning: It is a perfectly valid to only provide NetCDF with XIOS and skip HDF5.

From this I understand that HDF5 is not required. I will take care of updating the corresponding parts in the afternoon.

> Would it be safer to just move the existing case `IF(HAVE_XIOS)` ahead of the `IF(HAVE_NETCDF)`? (And maybe also fix the indentation of the `IF(HAVE_HDF5)` body while you are at it.)

Definitely! Since `TARGET_LINK_LIBRARIES` appends to the list of linked libraries this should simplify things a lot.

> If XIOS actually has a hard dependency on PHDF5 ...

Seems to be not the case. So no need for action here.

fem/src/CMakeLists.txt
BenjaminRodenberg: Will re-organize this part to avoid the nesting and just append to the list one-by-one.

fem/src/CMakeLists.txt
BenjaminRodenberg: See above. I will clean this up.

BenjaminRodenberg (commented):
Thank you for the review! I updated the code according to the suggested changes.

In the process I also decided to revert most of the changes in `elmerice/Solvers/CMakeLists.txt` because I could not trigger any failure and so I think it is not really justified to introduce larger changes. For consistency reasons the following change would still be possible but at least for the cases I'm looking at there seems to be no problem with the original version:

```diff
diff --git a/elmerice/Solvers/CMakeLists.txt b/elmerice/Solvers/CMakeLists.txt
index 4284af562..5ec779b23 100644
--- a/elmerice/Solvers/CMakeLists.txt
+++ b/elmerice/Solvers/CMakeLists.txt
@@ -152,17 +152,19 @@ IF(NOT(WIN32))
 ENDIF()
 TARGET_LINK_LIBRARIES(ElmerIceSolvers Elmer::MPI_Fortran elmersolver ElmerIceUtils)
 
-IF(HAVE_NETCDF)
-  TARGET_LINK_LIBRARIES(ElmerIceSolvers ${NETCDF_LIBRARIES})
-ENDIF()
-
-IF(HAVE_HDF5)
-  TARGET_LINK_LIBRARIES(ElmerIceSolvers ${PHDF5_LIBRARIES})
-ENDIF()
-
 IF(HAVE_XIOS)
   INCLUDE_DIRECTORIES(${XIOS_INCLUDE_DIR})
-  TARGET_LINK_LIBRARIES(ElmerIceSolvers ${XIOS_LIBRARIES})
+  TARGET_LINK_LIBRARIES(ElmerIceSolvers ${XIOS_LIBRARIES} ${NETCDF_LIBRARIES})
+  IF(HAVE_HDF5)
+    TARGET_LINK_LIBRARIES(ElmerIceSolvers ${PHDF5_LIBRARIES})
+  ENDIF()
+ELSE()
+  IF(HAVE_NETCDF)
+    TARGET_LINK_LIBRARIES(ElmerIceSolvers ${NETCDF_LIBRARIES})
+  ENDIF()
+  IF(HAVE_HDF5)
+    TARGET_LINK_LIBRARIES(ElmerIceSolvers ${PHDF5_LIBRARIES})
+  ENDIF()
 ENDIF()
 
 # IF(HAVE_MMG)
```

Another note: I'm currently only developing this change for the feature branch `yac-coupling` (planned to be merged into `devel` on the long run). I also want to check if the change is necessary for `devel` already now. If yes, we can also cherry-pick the relevant changes and merge this into `devel`.

BenjaminRodenberg:
I experimented with the changes from this PR and the current state of `devel`: If I configure with `elmerfem/build$ cmake -DWITH_NETCDF=ON -DWITH_XIOS=ON -DXIOS_ROOT=$XIOS_INSTALL ..` I get the same kind of behavior as for `yac-coupling`. Applying the changes here fixes this. From this perspective it would be justified to directly add the changes to `devel`. I used the commit 04db002d234cd1b84f341c97271e0152f496225e for testing and it should be possible to directly apply it to `devel`. (branch on my fork: [`devel-build-ubuntu`](https://github.com/BenjaminRodenberg/elmerfem/tree/devel-build-ubuntu))

Feel free to directly cherry-pick or I can also open another PR if you prefer this.

elmerice/Solvers/CMakeLists.txt
```diff
@@ -155,13 +155,16 @@ TARGET_LINK_LIBRARIES(ElmerIceSolvers Elmer::MPI_Fortran elmersolver ElmerIceUti
...
+
 IF(HAVE_HDF5)
-TARGET_LINK_LIBRARIES(ElmerIceSolvers ${PHDF5_LIBRARIES})
+  TARGET_LINK_LIBRARIES(ElmerIceSolvers ${PHDF5_LIBRARIES})
 ENDIF()
+
 IF(HAVE_XIOS)
   INCLUDE_DIRECTORIES(${XIOS_INCLUDE_DIR})
   TARGET_LINK_LIBRARIES(ElmerIceSolvers ${XIOS_LIBRARIES})
 ENDIF()
```
mmuetzel: Maybe, I missed something. But wasn't the main point of this PR that XIOS has a dependency on NetCDF?

Wouldn't that mean that it is necessary to move this block to before the block with `IF(HAVE_NETCDF)` (at least when using a linker like `ld.bfd`)? Is that now no longer necessary for some reason?

fem/src/CMakeLists.txt
```diff
@@ -254,7 +254,18 @@ IF(WITH_MPI)
     LIST(APPEND ELMERSOLVER_LIBRARIES Elmer::MPI_C ${Hypre_LIBRARIES})
   ENDIF()
 
-  IF(NETCDF_LIBRARIES)
+  IF(XIOS_LIBRARIES)
+    # XIOS depends on NetCDF (mandatory) and HDF5 (optional), since linking order matters, add all in one go
+    IF(NOT NETCDF_LIBRARIES)
+      message(FATAL_ERROR "XIOS_LIBRARIES is set but NETCDF_LIBRARIES is missing")
+    ENDIF()
```
mmuetzel: There is already a check for that in the the root CMakeLists.txt file:
https://github.com/ElmerCSC/elmerfem/blob/443cd5b2c07e5627fbc912094a73de4e50d21b3e/CMakeLists.txt#L493-L496

It is probably not necessary to duplicate that check here.

Or does the check in the root CMakeLists.txt file not trigger for some reason? In that case, that check should probably be fixed.

fem/src/CMakeLists.txt
```diff
@@ -254,7 +254,18 @@ IF(WITH_MPI)
...
+
+    LIST(APPEND ELMERSOLVER_LIBRARIES ${XIOS_LIBRARIES} ${NETCDF_LIBRARIES})
+    IF(PHDF5_LIBRARIES)
+      LIST(APPEND ELMERSOLVER_LIBRARIES ${PHDF5_LIBRARIES})
+    ENDIF()
```
mmuetzel: It is probably not necessary to duplicate the lines below here.
It would probably be easier to just move the original block starting with `IF(XIOS_LIBRARIES)` to before the block that links to the NetCDF libraries.

elmerice/Solvers/CMakeLists.txt
BenjaminRodenberg: I originally introduced the changes in `elmerice/Solvers/CMakeLists.txt` mainly for consistency reasons and later found out that they are not needed. The changes in `fem/src/CMakeLists.txt`, however, are definitely needed on my Ubuntu 24.04 System and compilation using the code from `devel` will fail if I configure with the following settings: 

```sh
elmerfem/build$ cmake -DWITH_NETCDF=ON -DWITH_XIOS=ON -DXIOS_ROOT=$XIOS_INSTALL ..
```

I do not observe this behavior on Levante and I assume on LUMI it's also working when using [this script](https://github.com/ElmerCSC/elmerfem/blob/443cd5b2c07e5627fbc912094a73de4e50d21b3e/ci/LUMI/elmer_install_gcc_spack.sh#L110)? For Ubuntu XIOS is not active in the CI (see [here](https://github.com/ElmerCSC/elmerfem/blob/443cd5b2c07e5627fbc912094a73de4e50d21b3e/.github/workflows/ubuntu-gcc-full.yaml#L52-L71)).

fem/src/CMakeLists.txt
BenjaminRodenberg: The check in the CMakeLists.txt works perfectly fine. Adding another check here makes this requirement more explicit but I can also remove it if you prefer.

fem/src/CMakeLists.txt
BenjaminRodenberg: I'm also not a big fan of duplication. My intention here was to get a self contained `IF(XIOS_LIBRARIES)`-block and for this I would be willing to accept some code duplication. With the current solution it is harder to break the linking order of HDF5, NetCDF and XIOS. If the HDF5&NetCDF block are independent from the XIOS block I can imagine that through some reordering this could happen again in the future and we end up with the same problem.

fem/src/CMakeLists.txt
```diff
@@ -265,10 +276,6 @@ IF(WITH_MPI)
     LIST(APPEND ELMERSOLVER_LIBRARIES ${FETI4I_LIBRARIES})
   ENDIF()
 
-  IF(XIOS_LIBRARIES)
-    LIST(APPEND ELMERSOLVER_LIBRARIES ${XIOS_LIBRARIES})
-  ENDIF()
-
```
BenjaminRodenberg: Shifting this block before the `IF(NETCDF_LIBRARIES)` would be the alternative with less code duplication but from my perspective also slightly less safe with respect to future risk of breaking. We can also go for this option if you prefer.

BenjaminRodenberg (commented):
I added a few more comments to clarify my motivation for the changes.

Generally, I have the impression that adding XIOS to the CI for Ubuntu would also help a lot to ensure everything works on Ubuntu - especially since I did not see any trouble building on Levante. But I cannot really judge if this is something you want in your CI because I assume XIOS adds quite some building time.

What do you think about directly adding these changes to `devel` without the detour over `yac-coupling`? Potentially with some additional CI checks?

elmerice/Solvers/CMakeLists.txt
mmuetzel: I still think that it would be best to re-order this in the following order:
* XIOS
* NETCDF
* HDF5

Maybe also leave a comment why this order is needed (dependencies).

fem/src/CMakeLists.txt
mmuetzel: I'm not a big fan of dead code that can never be reached. So, I'd be in favor of removing this check here.

fem/src/CMakeLists.txt
mmuetzel: I'd be in favor of that (more simple) change. Maybe, leave a comment why the XIOS block needs to be before the NETCDF (and HDF5) block.

mmuetzel:
Thank you for the follow up changes. If this still works for your use-case, this looks good to me. 👍

I think these changes should go to the `devel` branch, too. Could you please update the target branch?

BenjaminRodenberg:
Thanks for the review! I applied the suggested changes. I also updated my branch `devel-build-ubuntu`: https://github.com/ElmerCSC/elmerfem/compare/devel...BenjaminRodenberg:elmerfem:devel-build-ubuntu. I would prefer to merge the changes into `devel` if possible. We are overdue merging `devel` back into `yac-coupling` anyhow.

BenjaminRodenberg:
Updated this branch and the target :+1: Previously, it contained a lot of changes from `yac-coupling` that are not ready for `devel` yet. Therefore, had to overwrite `yac-coupling-build-ubuntu` via force push. Hope this is fine.

raback:
So once these tests have passed I guess this can be merged? In ~10 minutes or so....

PR #743, opened 2026-01-06
Add prefix to remove potential ambiguity

BenjaminRodenberg:
* Add prefix elmer_coupling_* to avoid ambiguity for subroutine mpi_handshake
* Ambiguity led to failure if mpi_handshake from XIOS is used in SParIterComm while mpi_handshake from YAC is used in elmer_coupling

This should fix the bug which made the work-around introduced in https://github.com/ElmerCSC/elmerfem/commit/aca13887bdf37de0d6d9aca83022c8c021fe2965 necessary. Both mpi_handshake implementations from YAC and XIOS should now work equally fine.


BenjaminRodenberg:
I still want to do some manual testing on Levante to ensure that the problem is indeed gone. Therefore, marking this as draft.

BenjaminRodenberg:
@ClemensSchannwell : I'm adding a patch that swaps the order of YAC and XIOS when the handshake is imported. Giving priority to the mpi_handshake from XIOS was not possible without the fix [427340b](https://github.com/ElmerCSC/elmerfem/pull/743/commits/427340b302608dbabca2796adc2943fe01cffe5c). It should now work without problems.

Please apply the patch via `git apply swap_yac_xios_mpi_handshake_precedence.patch` locally to test this PR.

[swap_yac_xios_mpi_handshake_precedence.patch](https://github.com/user-attachments/files/24818836/swap_yac_xios_mpi_handshake_precedence.patch)

Thanks for helping out with the testing and let me know if you need help or something looks odd!

BenjaminRodenberg:
If you are wondering about the mess: I did a bad merge of `BenjaminRodenberg/yac-coupling` into this branch. `BenjaminRodenberg/yac-coupling` includes `ElmerCSC/devel`, but `ElmerCSC/yac-coupling` (which this PR targets) does not include these changes (yet). So this would result in a gigantic and unwanted diff. Therefore, I had to revert this.

The patch from above still applies.

BenjaminRodenberg:
Ok. I merged `devel` into `yac-coupling` and then the (updated) `yac-coupling` into this PR. Now we should be fine :+1:

ClemensSchannwell:
So I tested the patch on [ebfm-elmer-icon](https://github.com/ClemensSchannwell/elmerfem/tree/ebfm-elmer-icon)  and could successfully run our testcase as well as 1-day test simulation of ICON-EBFM-Elmer.  Seems to have solved the previously present problem.

PR #745, opened 2026-01-08
Fix building with quadruple-precision floating-point support

mmuetzel:
This essentially the same as #739 but doesn't add the CI runner using Flang.

I guess it is easier to merge just the parts that fix support for quadruple-precision floating-support first.
I'll rebase the other PR on top of this one (in case this PR is acceptable).


raback:
Thanx! It seems that my quick fix, came with some errors. One should learn that there is not so small change that it could not have a bug...

PR #747, opened 2026-01-13
Overhaul documentation for building Elmer on Windows using MSYS2

mmuetzel:


raback:
Thanx! This may be helpful also as a recipe when updating the VM that creates the nightly builds. It also deserves an overhaul.

PR #754, opened 2026-01-22
Add fallback value for `ELMER_FEM_BRANCH`

mmuetzel:
If building from a detached head or from a source tree that is not a Git repository, `git branch --show-current` fails.

Provide a fallback value for `ELMER_FEM_BRANCH` to avoid compilation errors like:
```
/home/runner/work/elmerfem/elmerfem/elmergrid/src/fempre.c:86:71: error: ‘ELMER_FEM_BRANCH’ undeclared (first use in this function)
   86 |   printf("Version: %s-%s (Rev: %s, Compiled: %s)\n",ELMER_FEM_VERSION,ELMER_FEM_BRANCH,&
      |                                                                       ^~~~~~~~~~~~~~~~
/home/runner/work/elmerfem/elmerfem/elmergrid/src/fempre.c:86:71: note: each undeclared identifier is reported only once for each function it appears in

```


mmuetzel:
Wait a second: What is the intension for `RESET_ELMER_BRANCH`?
It is a cache variable that can be set by a user.
If a user configures with `-DRESET_ELMER_BRANCH=OFF`, it would still fail (unless the user manually sets `-DELMER_FEM_BRANCH="something"`).

raback:
I committed another quick fix already for ElmerGrid side. ElmerSolver already checked for the existance. I am not any Cmake wizard so I copy-pasted the logic from the revision using just different git command. So I am sure they could be made prettier. Just thought that it would be nice to know what is the current branch.

mmuetzel:
Your change looks like it should avoid the error. 👍

PR #755, opened 2026-01-23
Add accessor function for `listheaders` with OpenMP.

mmuetzel:
Accessing data with thread-local storage doesn't work reliably on Windows.

Avoid direct access to `listheaders` by using an accessor function if MATC is built with OpenMP.

This should allow enabling MATC together with OpenMP when using a LLVM toolchain for Windows targets.

I'm not sure what `ld.bfd` did to make that kind of work. But `ld.lld` fails to link ElmerGUI on Windows without this change.
Maybe, that is also the reason for the tests that are failing randomly in CI? (Probably not. Afaict, this change only affects ElmerGUI.)


mmuetzel:
> Maybe, that is also the reason for the tests that are failing randomly in CI? (Probably not. Afaict, this change only affects ElmerGUI.)

I guess the CI answered that question: There are still intermittent test failures with this change.

juharu:
Thanks Markus!

PR #757, opened 2026-01-26
Add OpenMP include directories to compiler and preprocessor flags

mmuetzel:
Some implementations of OpenMP (e.g., LLVM libomp) require that the directory where the `omp_lib` module is installed is explicitly passed as a compiler/preprocessor flag.

Ideally that should be done by linking all CMake targets that use OpenMP with the respective OpenMP CMake target.
For the time being, add the directory manually (to all targets).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -207,6 +207,9 @@ IF(WITH_OpenMP)
   SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}")
   SET(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} ${OpenMP_Fortran_FLAGS}")
   SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
+  IF(OpenMP_Fortran_INCLUDE_DIR)
+    INCLUDE_DIRECTORIES("${OpenMP_Fortran_INCLUDE_DIR}")
+  ENDIF()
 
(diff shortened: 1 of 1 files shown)


mmuetzel:
See https://github.com/ElmerCSC/elmerfem/pull/753#issuecomment-4651401314 where this change seems to have helped with a build issue when enabling OpenMP with disabled MPI. In that case, `flang` wasn't able to locate the `omp_lib.mod` module. The build errors without this change looked like this:
```
D:\a\_temp\msys64\clang64\bin\flang.exe -ID:\a\elmerfem\elmerfem\fem\src\modules -ID:/a/elmerfem/elmerfem/build -ID:/a/elmerfem/elmerfem/build/fem/src -ID:/a/elmerfem/elmerfem/fhutiter/include -fopenmp -O3 -Jfmodules -ffixed-line-length-72 -c fem/src/modules/CMakeFiles/UMATLib.dir/UMATLib.F90-pp.f90 -o fem/src/modules/CMakeFiles/UMATLib.dir/UMATLib.F90.obj
error: Semantic errors in fem/src/modules/CMakeFiles/UMATLib.dir/UMATLib.F90-pp.f90
fmodules/types.mod:82:5: error: Cannot parse module file for module 'omp_lib': Source file 'omp_lib.mod' was not found
  use omp_lib,only:omp_integer_kind
      ^^^^^^^
fmodules/types.mod:83:5: error: Cannot parse module file for module 'omp_lib': Source file 'omp_lib.mod' was not found
  use omp_lib,only:omp_logical_kind
      ^^^^^^^
[...]
```

PR #759, opened 2026-01-26
Increase stack size for ElmerSolver executable on Windows to 8 MB

mmuetzel:
The default stack size of 1 MB on Windows is very small for scientific programs like Elmer. Increase it to 8 MB.


juharu:
Thanks Markus!

PR #760, opened 2026-01-26
Read command line arguments in executable

mmuetzel:
The Flang runtime on Windows is static. That means that there are separate instances of the Flang runtime in each executable and DLL. The command line arguments are stored in the Flang runtime. Since they are different instances, the Flang runtime in the DLL has no way of knowing the command line arguments of the executable.
That looks like a design issue of the Flang runtime on Windows.

Work around that by collecting the command line arguments in the context of the executable and passing them to the entry point (`ElmerSolver`) of the DLL.

### DIFF
--- a/fem/src/ElmerSolver.F90
+++ b/fem/src/ElmerSolver.F90
@@ -54,7 +54,7 @@
 !------------------------------------------------------------------------------
 !> The main program for Elmer. Solves the equations as defined by the input files.
 !------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


mmuetzel:
This worked for me in local tests in the CLANG64 environment of MSYS2.
But my Fortran is pretty rusty. Please, check if this is actually ok.

juharu:
Nice, thanks!

PR #761, opened 2026-01-27
Remove unsafe OpenMP parallel region around Lua evaluation

mmuetzel:
The inner `!$OMP PARALLEL` section in `TrimLuaExpression` evaluated a single Lua expression in multiple threads while sharing `lua_result`, `result_len`, and several loop indices. This caused data races and intermittent segmentation faults, especially with LLVM Flang/MinGW.

Fix this by removing the inner parallel section and serializing the Lua evaluation.

I'm not an OpenMP expert. But I think this fixes an actual issue.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -171,6 +171,7 @@ jobs:
         timeout-minutes: 150
         env:
           CTEST_OUTPUT_ON_FAILURE: 1
+          OMP_NUM_THREADS: 4
         run: |
           cd ${GITHUB_WORKSPACE}/build
           ctest . \
--- a/fem/src/GeneralUtils.F90
+++ b/fem/src/GeneralUtils.F90
@@ -1242,9 +1242,7 @@ CONTAINS
              closed_region = .FALSE.
            END IF
 
-           !$OMP PARALLEL DEFAULT(NONE) &
-           !$OMP FIRSTPRIVATE(tcmdstr, tninlen, lstat) &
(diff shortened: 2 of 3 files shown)


juhanikataja:
That shouldn't work since now at `# <lua_code>` regions the `<lua_code>` is not executed for all instances of LuaState. I made lua test more strict about this, but the error is not caught if the ci tests don't run with multiple threads.

juhanikataja:
The original code should be correct. The shared variables are not modified in the parallel region and if they are, they are inside critical region where they are executed only one thread at a time (the critical region).

mmuetzel:
If I understand correctly, `lua_result` and `result_len` (and maybe others) can't be shared variables. Otherwise one thread might override their values for the other threads before they are actually used.

mmuetzel:
I'm not sure if the test are run with multiple OpenMP threads by default. (I seem to recall that logic was added to use only one thread unless `OMP_NUM_THREADS` is set in the environment.)
I pushed a change that sets that variable to 4 for the Ubuntu runners as a test.

mmuetzel:
I didn't understand what is supposed to be happening. (And I'm still not sure I do.)
But your change clearly shows that this PR is wrong.

Sorry for the noise.

juhanikataja:
No problem! Thanks for pointing out that tests are ran with 1 thread. We should really run threaded tests with more than 1 threads...

mmuetzel:
> We should really run threaded tests with more than 1 threads...

#740 was an attempt at doing that. 
But one test failed at that time. And I assume that was the reason that is on hold currently.

PR #763, opened 2026-01-28
CI: Fix typo in configure flags for Ubuntu-parallel workflow.

mmuetzel:
Currently, the rules build without ElmerIce.
Fix spelling of flag `WITH_ElmerIce` (CMake flags are case-sensitive).


juharu:
Thanks!

Issue #723, opened 2025-11-24
YAC coupling: mesh coordinates from Elmer internals


ClemensSchannwell:
In the `yac-coupling` branch, where we couple ICON and the EBFM to Elmer/Ice, the elmer grid information that `YAC` needs to interpolate the fields is currently computed using an external c-function (https://github.com/ElmerCSC/elmerfem/blob/yac-coupling/fem/src/elmer_grid.c). This function reads in the coordinates from the partitioned mesh files. The `proj-string` is currently also hard-coded (https://github.com/ElmerCSC/elmerfem/blob/yac-coupling/fem/src/elmer_grid.c#L59-L61). While this is working for the moment, it makes running Elmer/Ice on different domains (e.g. Svalbard, Antarctica) cumbersome for now. 
From my point of view it would be ideal, if the user can provide the `proj-string`, while the mesh coordinates, cell-centers etc. are used from Elmer internals.

A shopping list of what `YAC` needs is the following:
```
int num_nodes // number of nodes stored on the local process
int num_elem // number of elements stored on the local process
int node_ids[num_nodes] // global ids of all local nodes
int elem_ids[num_elem] // global ids of all elem
double node_x[num_nodes] // x coords of local nodes
double node_y[num_nodes] // y coords  of local nodes
int num_node_per_elem[num_elem] or int num_node_per_elem // number of nodes for each elem
int elem_to_node[SUM(num_node_per_elem[:])] or int elem_to_node[num_node_per_elem * num_elem] // local index of nodes for each elem; node of each elem have to be order either clock or counter-clock wise; values are in the range [0;num_nodes-1]; nodes of i'th elem are stored at elem_to_node[SUM(num_node_per_elem[0:i-1]):SUM(num_node_per_elem[0:i])]

some kind of information on the projection being used

If available (I can also compute it from node_x/node_y):
double elem_x[num_elem] // x coords of center of elem
double elem_y[num_elem] // y coords of center of elem
```

raback:
I sent a suggested code directly as it was still just a template.

BenjaminRodenberg:
I will start working on this issue on the following branch of my fork: ~~https://github.com/BenjaminRodenberg/elmerfem/tree/yac-coupling-direct-mesh~~ https://github.com/BenjaminRodenberg/elmerfem/pull/1

fgillet:
Hi don't know if it can be useful for you; but for the coupling with XIOS I also need to send the grid information to xios; 
[...]

PR #764, opened 2026-01-28
Remove `USE Messages` from `MODULE Types`

BenjaminRodenberg:
Remove implicit re-export of `Messages` through `Types` module and add explicit `USE Messages, ONLY: <symbols>` to all affected files. This clarifies dependencies and prepares for upcoming YAC coupling integration, which would otherwise create a circular dependency (`Messages -> elmer_coupling -> Types -> Messages`).

See https://github.com/BenjaminRodenberg/elmerfem/pull/1 for an example where the circular dependency caused problems. https://github.com/BenjaminRodenberg/elmerfem/pull/1 is a draft for https://github.com/ElmerCSC/elmerfem/issues/723.

### DIFF
--- a/fem/src/Messages.F90
+++ b/fem/src/Messages.F90
@@ -55,6 +55,8 @@ MODULE Messages
   USE XIOS
 #endif
 
+   USE Types_  ! Check for circular dependencies
+
    IMPLICIT NONE
    
    CHARACTER(LEN=512) :: Message = ' '
--- a/fem/src/PElementMaps.F90
+++ b/fem/src/PElementMaps.F90
@@ -47,6 +47,7 @@
 
 MODULE PElementMaps
   USE Types
+  USE Messages, ONLY: Info, Warn, Fatal, Message
   Use GeneralUtils, ONLY : I2S
 
   IMPLICIT NONE
--- a/fem/src/CMakeLists.txt
+++ b/fem/src/CMakeLists.txt
@@ -25,9 +25,9 @@ SET(solverlib_SOURCES AddrFunc.F90 NavierStokes.F90 NavierStokesGeneral.F90
   LinearAlgebra.F90 CoordinateSystems.F90 ListMatrix.F90 CRSMatrix.F90
   BandMatrix.F90 BandwidthOptimize.F90 BlockSolve.F90
   MaterialModels.F90 DirectSolve.F90 IterSolve.F90
-  IterativeMethods.F90 TimeIntegrate.F90 Types.F90 SolveBand.F90
-  ElementUtils.F90 Radiation.F90 fft.c Load.c Differentials.F90
-  FreeSurface.F90 
+  IterativeMethods.F90 TimeIntegrate.F90 Types.F90 TypesWrapper.F90
+  SolveBand.F90 ElementUtils.F90 Radiation.F90 fft.c Load.c Differentials.F90
+  FreeSurface.F90
   Walls.F90 SolverUtils.F90 SolveSBand.F90 CPUTime.c Interpolation.F90
   MainUtils.F90 Adaptive.F90 EigenSolve.F90 HashTable.F90 MortarUtils.F90
   MeshAllocations.F90 GeometryFitting.F90
--- a/fem/src/DirectSolve.F90
+++ b/fem/src/DirectSolve.F90
@@ -48,6 +48,8 @@
 
 MODULE DirectSolve
 
+   USE Messages, ONLY: EXIT_ERROR
+
    USE CRSMatrix
    USE BandMatrix
    USE SParIterSolve
--- a/fem/src/EigenSolve.F90
+++ b/fem/src/EigenSolve.F90
@@ -57,6 +57,8 @@
 
 MODULE EigenSolve
 
+   USE Messages, ONLY: Info, Error, Fatal
+
    IMPLICIT NONE
 
 CONTAINS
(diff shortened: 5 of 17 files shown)


BenjaminRodenberg:
Please let me know if you expect a specific kind of ordering for imports. I first wanted to make sure this change goes in the right direction before taking care of such details. Hope this is ok.

fem/src/DefUtils.F90
```diff
@@ -3471,7 +3471,7 @@ RECURSIVE SUBROUTINE DefaultSlaveSolvers( Solver, SlaveSolverStr)
 
      INTERFACE
        SUBROUTINE SolverActivate_x(Model,Solver,dt,Transient)
-         USE Types
+         USE Types_
```
BenjaminRodenberg: Not sure if this could cause trouble on the user side? `CompressibleNS.F90`, for example uses `DefUtils`. I could also add a `DefUtilsWrapper.F90` similar to `TypesWrapper.F90` but would happier to not do this.

I need this change since otherwise my code in https://github.com/BenjaminRodenberg/elmerfem/pull/1 will not compile due to [this part](https://github.com/BenjaminRodenberg/elmerfem/blob/d3d9b3311322fc079089aa35bb46ec3682953b52/elmerice/Solvers/yac2elmer.F90#L2-L7):

```
  USE DefUtils, ONLY: ParEnv, GetSolverParams, GetString, GetMesh, GetNOFActive, DefaultVariableAdd
  USE SolverUtils
  USE elmer_coupling, ONLY: coupling_setup, is_main_rank
  USE elmer_ebfm_coupling, ONLY: elmer_ebfm_interface, t_ice_field, smb_field, &
                                 runoff_field
  ! USE elmer_icon_coupling, ONLY: elmer_icon_interface, clt_field
```

`DefUtils` implicitly imports `Messages` if using `Types`.`Fatal` from `Messages` calls `coupling_finalize`. This conflicts with importing `elmer_coupling` since it also relies on `Types` and we get the circular dependency.

BenjaminRodenberg:
To mimic the changes from https://github.com/BenjaminRodenberg/elmerfem/pull/1 that motivate this PR I added the following part here:

```
diff --git a/fem/src/Messages.F90 b/fem/src/Messages.F90
index 8a547b297..d22c71642 100644
--- a/fem/src/Messages.F90
+++ b/fem/src/Messages.F90
@@ -55,6 +55,8 @@ MODULE Messages
   USE XIOS
 #endif
 
+   USE Types_  ! Check for circular dependencies
+
    IMPLICIT NONE
    
    CHARACTER(LEN=512) :: Message = ' '
```

When trying to build this modified version I also had to change the build order (see 18cd4fb). But this was generally quite hard to reproduce and I think this problem also only shows up if building with `-DCMAKE_BUILD_TYPE=DEBUG`. It might also be due to a race condition with parallel builds...

fem/src/Messages.F90
```diff
@@ -55,6 +55,8 @@ MODULE Messages
   USE XIOS
 #endif
 
+   USE Types_  ! Check for circular dependencies
```
BenjaminRodenberg, line 58: This is not possible if `Types_` (formerly `Types`) already includes `Messages`.

`USE :: Types_, ONLY: Mesh_t, Element_t, dp` is needed by `MODULE elmer_coupling` imported in `Messages` (https://github.com/BenjaminRodenberg/elmerfem/pull/1).

fem/src/PElementMaps.F90
```diff
@@ -47,6 +47,7 @@
 
 MODULE PElementMaps
   USE Types
+  USE Messages, ONLY: Info, Warn, Fatal, Message
```
BenjaminRodenberg, line 50: Explicitly importing `Messages` is needed if using `Types_` (without `Messages`) instead of `Types`.

BenjaminRodenberg (commented):
After going a bit back and forth I'm reaching a point in this PR where things could be merged.

The main open question is whether a wrapper module `Types` should be introduced for backwards compatibility. Without this module, tests will fail at the moment because some also rely on implicitly importing subroutines from `Messages` which were remove from the original `Types` (which is renamed to `Types_` here).

I'm not a big fan of providing such a wrapper module because it supports an anti-pattern. I would rather push users to explicitly `USE Messages`. It's hard to judge from my perspective whether this is considered a breaking change or a bugfix.

I think it generally does not harm to explicitly import the subroutines from `Messages` so I kept this part.

Note: The build order was no problem and due to a misconfiguration of my build workflow.

BenjaminRodenberg:
I squashed most commits in this PR. [1c826ce](https://github.com/ElmerCSC/elmerfem/pull/764/commits/1c826cec1bfede55f8539ea1cbcadf59593846ae) is only for testing and should be reverted before the merge.

BenjaminRodenberg:
Discussed this with @tzwinger. Here is a summary about the current status of this PR from my side:

* The alternative solution provided on the branch [`SeraparateMessagesModule`](https://github.com/ElmerCSC/elmerfem/tree/SeraparateMessagesModule) works for me. I tested the integration on my fork on the branch [`yac-coupling-no-Wrapper`](https://github.com/BenjaminRodenberg/elmerfem/tree/yac-coupling-no-Wrapper).
* I had to move some functionality around in #783. This also allowed me to remove the import of `Types` from my code and I do not run into the cyclic dependency anymore if `USE Messages` is called inside `Types`. I still think that the way how `Messages` is imported implicitly at the moment is not ideal and limits flexibility. From that perspective it's good if this is changed - but the current state is also not problematic anymore.

I would suggest to close this PR in favor of the changes on [`SeraparateMessagesModule`](https://github.com/ElmerCSC/elmerfem/tree/SeraparateMessagesModule). If we want to change something about how `Types` and `Messages` are imported this would rather be moving into the direction of [`SeraparateMessagesModule`](https://github.com/ElmerCSC/elmerfem/tree/SeraparateMessagesModule).

raback:
I merged this branch to "devel" so everything should be ok
https://github.com/ElmerCSC/elmerfem/tree/SeraparateMessagesModule
Note that I did a squash to combine the 7 baby steps so rather merge to devel than this. 
I will close this PR since to my understanding it is no longer needed.

PR #766, opened 2026-02-03
Skip CI job `ubuntu-parallel` on forks

BenjaminRodenberg:
When developing on my fork the test `ubuntu-parallel` always fails due to not having a self-hosted runner. I'm not sure why you are using a self-hosted runner but with some modifications one can also use `ubuntu-latest` and fix the test for forks.

If you really need a self-hosted runner: One could introduce a case distinction. (either deactivate test for forks or use `self-hosted` on `ElmerCSC/elmerfem` and `ubuntu-latest` else)

### DIFF
--- a/.github/workflows/ubuntu-parallel.yaml
+++ b/.github/workflows/ubuntu-parallel.yaml
@@ -14,7 +14,7 @@ on:
 jobs:
 
   ubuntu:
-    if: ${{ github.repository == 'ElmerCSC/elmerfem' }}  # only run job for main repository (forks usually not have self-hosted runners)
+    if: ${{ github.repository == 'ElmerCSC/elmerfem' }}  # only run job for main repository (forks usually cannot access the "self-hosted" runner)
     runs-on: self-hosted
 
     name: ubuntu-parallel (${{ matrix.compiler }})
--- a/.github/workflows/ubuntu-parallel.yaml
+++ b/.github/workflows/ubuntu-parallel.yaml
@@ -14,7 +14,7 @@ on:
 jobs:
 
   ubuntu:
-    if: ${{ github.repository == 'ElmerCSC/elmerfem' }}  # only run job for main repository (forks usually cannot access the "self-hosted" runner)
+    if: ${{ github.repository == 'ElmerCSC/elmerfem' }}  # only run job for main repository (forks usually do not have self-hosted runners)
     runs-on: self-hosted
 
     name: ubuntu-parallel (${{ matrix.compiler }})
--- a/.github/workflows/ubuntu-parallel.yaml
+++ b/.github/workflows/ubuntu-parallel.yaml
@@ -71,7 +71,7 @@ jobs:
             -DCREATE_PKGCONFIG_FILE=ON \
             -DWITH_MPI=ON \
             -DMPI_TEST_MAXPROC=$(nproc) \
-            -DMPIEXEC_PREFLAGS="--allow-run-as-root --oversubscribe" \
+            -DMPIEXEC_PREFLAGS="--allow-run-as-root;--oversubscribe" \
             ..
 
       - name: build
--- a/.github/workflows/ubuntu-parallel.yaml
+++ b/.github/workflows/ubuntu-parallel.yaml
@@ -14,7 +14,14 @@ on:
 jobs:
 
   ubuntu:
-    runs-on: ${{ github.repository == 'ElmerCSC/elmerfem' && 'self-hosted' || 'ubuntu-latest' }}
+    runs-on: |-
+      # Conditional runner selection:
+      # - ElmerCSC/elmerfem repository -> self-hosted
+      # - Forks (default) -> ubuntu-latest
+      ${{ case(
+        github.repository == 'ElmerCSC/elmerfem', 'self-hosted',
+        'ubuntu-latest'                                       
+      ) }}
     
     name: ubuntu-parallel (${{ matrix.compiler }})
 
--- a/.github/workflows/ubuntu-parallel.yaml
+++ b/.github/workflows/ubuntu-parallel.yaml
@@ -72,7 +72,7 @@ jobs:
             -DCREATE_PKGCONFIG_FILE=ON \
             -DWITH_MPI=ON \
             -DMPI_TEST_MAXPROC=$(nproc) \
-            -DMPIEXEC_PREFLAGS="--allow-run-as-root\;--oversubscribe" \
+            -DMPIEXEC_PREFLAGS="--allow-run-as-root" \
             ..
 
       - name: build
--- a/.github/workflows/ubuntu-parallel.yaml
+++ b/.github/workflows/ubuntu-parallel.yaml
@@ -71,7 +71,7 @@ jobs:
             -DCREATE_PKGCONFIG_FILE=ON \
             -DWITH_MPI=ON \
             -DMPI_TEST_MAXPROC=$(nproc) \
-            -DMPIEXEC_PREFLAGS="--allow-run-as-root;--oversubscribe" \
+            -DMPIEXEC_PREFLAGS="--allow-run-as-root\;--oversubscribe" \
             ..
 
       - name: build
--- a/.github/workflows/ubuntu-parallel.yaml
+++ b/.github/workflows/ubuntu-parallel.yaml
@@ -14,15 +14,9 @@ on:
 jobs:
 
   ubuntu:
-    runs-on: |-
-      # Conditional runner selection:
-      # - ElmerCSC/elmerfem repository -> self-hosted
-      # - Forks (default) -> ubuntu-latest
-      ${{ case(
-        github.repository == 'ElmerCSC/elmerfem', 'self-hosted',
-        'ubuntu-latest'                                       
-      ) }}
-    
+    if: ${{ github.repository == 'ElmerCSC/elmerfem' }}  # only run job for main repository (forks usually not have self-hosted runners)
+    runs-on: self-hosted
+
     name: ubuntu-parallel (${{ matrix.compiler }})
 
     strategy:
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -104,7 +104,7 @@
 
                 "-DWITH_ElmerIce:BOOL=TRUE"
 
-                "-DMPIEXEC_PREFLAGS=-oversubscribe"
+                "-DMPIEXEC_PREFLAGS=--oversubscribe"
```
BenjaminRodenberg, line 107: This is not 100% related to this PR but I stumbled across this while working on the PR.

It's apparently only cosmetics since (on my laptop) `--oversubscribe` and `-oversubscribe` do the same.

```
$ mpirun -n 15 -oversubscribe echo "Helloworld!"
Helloworld!
Helloworld!
Helloworld!
Helloworld!
Helloworld!
Helloworld!
Helloworld!
Helloworld!
Helloworld!
Helloworld!
Helloworld!
Helloworld!
Helloworld!
Helloworld!
Helloworld!
```

I would prefer `--oversubscribe`, since the error if you do not provide the option also uses two dashes:

```
$ mpirun -n 15 echo "Helloworld!"
--------------------------------------------------------------------------
There are not enough slots available in the system to satisfy the 15
slots that were requested by the application:

  echo

Either request fewer slots for your application, or make more slots
available for use.

A "slot" is the Open MPI term for an allocatable unit where we can
launch a process.  The number of slots available are defined by the
environment in which Open MPI processes are run:

  1. Hostfile, via "slots=N" clauses (N defaults to number of
     processor cores if not provided)
  2. The --host command line parameter, via a ":N" suffix on the
     hostname (N defaults to 1 if not provided)
  3. Resource manager (e.g., SLURM, PBS/Torque, LSF, etc.)
  4. If none of a hostfile, the --host command line parameter, or an
     RM is present, Open MPI defaults to the number of processor cores

In all the above cases, if you want Open MPI to default to the number
of hardware threads instead of the number of processor cores, use the
--use-hwthread-cpus option.

Alternatively, you can use the --oversubscribe option to ignore the
number of available slots when deciding the number of processes to
launch.
--------------------------------------------------------------------------
```

.github/workflows/ubuntu-parallel.yaml
```diff
@@ -71,7 +71,7 @@ jobs:
             -DCREATE_PKGCONFIG_FILE=ON \
             -DWITH_MPI=ON \
             -DMPI_TEST_MAXPROC=$(nproc) \
```
mmuetzel, line 74: Switching from the self-hosted runner to a GitHub-hosted runner would reduce the number of tests by this workflow by a lot unless you manually set this to a high value. E.g.:
``` suggestion
            -DMPI_TEST_MAXPROC=16 \
```

mmuetzel:
If the main motivation is to not have a workflow that never starts on a fork:
You could also disable this workflow on your fork. Or maybe easier for anyone forking: the workflow conditions could be modified so that it only runs on this repository.

.github/workflows/ubuntu-parallel.yaml
mmuetzel: IIUC, having those tests with high number of MPI processes is the main motivation for that workflow in the first place.

BenjaminRodenberg:
Thanks for the context. This helps a lot. Would the following solution could be a good compromise?

* On fork run it as it is in this PR (oversubscribe and `ubuntu-latest`)
* On ElmerSCS continue running on `self-hosted` to avoid skipping tests (oversubscribe does not harm. So I would keep it for sake of simplicity)

.github/workflows/ubuntu-parallel.yaml
```diff
@@ -14,7 +14,14 @@ on:
...
   ubuntu:
-    runs-on: self-hosted
+    runs-on: |-
+      # Conditional runner selection:
+      # - ElmerCSC/elmerfem repository -> self-hosted
+      # - Forks (default) -> ubuntu-latest
+      ${{ case(
+        github.repository == 'ElmerCSC/elmerfem', 'self-hosted',
+        'ubuntu-latest'                                       
+      ) }}
```
BenjaminRodenberg: [`case`](https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#case) is a relatively new feature but I think it's much easier to read (and extend) than [`${{ x && 'ifTrue' || 'ifFalse' }}`](https://github.com/actions/runner/issues/409#issuecomment-752775072).

mmuetzel:
I'm not sure if any tests are run for this workflow that aren't already run in other workflows if you use a GitHub-hosted runner without manually changing `MPI_TEST_MAXPROC=$(nproc)`.
So, it might as well be skipped entirely in forks to save some CPU cycles (and energy).

.github/workflows/ubuntu-parallel.yaml
```diff
@@ -71,7 +72,7 @@ jobs:
             -DCREATE_PKGCONFIG_FILE=ON \
             -DWITH_MPI=ON \
             -DMPI_TEST_MAXPROC=$(nproc) \
-            -DMPIEXEC_PREFLAGS="--allow-run-as-root" \
+            -DMPIEXEC_PREFLAGS="--allow-run-as-root\;--oversubscribe" \
```
BenjaminRodenberg: If we skip the test for forks this should not be necessary (and maybe you even want to provoke a failure on the self-hoster runner if there are not sufficiently many cores).

So I would actually tend to reverting this change.

```suggestion
            -DMPIEXEC_PREFLAGS="--allow-run-as-root" \
```

flake.nix
BenjaminRodenberg: Still does not harm.

BenjaminRodenberg (commented):
Skipping the test sounds also good to me and makes things much simpler.

.github/workflows/ubuntu-parallel.yaml
```diff
@@ -14,8 +14,9 @@ on:
 jobs:
 
   ubuntu:
+    if: ${{ github.repository == 'ElmerCSC/elmerfem' }}  # only run job for main repository (forks usually not have self-hosted runners)
```
mmuetzel: nit
``` suggestion
    if: ${{ github.repository == 'ElmerCSC/elmerfem' }}  # only run job for main repository (forks usually cannot access the "self-hosted" runner)
```

"self-hosted" is the identifier for the runner that the Elmer maintainers set up. It isn't a generic keyword for any self-hosted runner afaik.

BenjaminRodenberg: As far as I know `self-hosted` is a common github-actions term. See [here](https://docs.github.com/en/actions/how-tos/manage-runners/self-hosted-runners/use-in-a-workflow#using-default-labels-to-route-jobs). Technically I could also set up a self-hosted runner for my fork (I assume few people actually do this):

image

.github/workflows/ubuntu-parallel.yaml
```diff
@@ -14,8 +14,9 @@ on:
 jobs:
 
   ubuntu:
+    if: ${{ github.repository == 'ElmerCSC/elmerfem' }}  # only run job for main repository (forks usually cannot access the "self-hosted" runner)
```
BenjaminRodenberg:
```suggestion
    if: ${{ github.repository == 'ElmerCSC/elmerfem' }}  # only run job for main repository (forks usually cannot access the self-hosted runner)
```
Maybe without the `"` but rest is also fine.

.github/workflows/ubuntu-parallel.yaml
mmuetzel: Thanks for clarifying. TIL
In that case, just adding a "do" before the "not" is probably sufficient.

.github/workflows/ubuntu-parallel.yaml
BenjaminRodenberg:
```suggestion
    if: ${{ github.repository == 'ElmerCSC/elmerfem' }}  # only run job for main repository (forks usually do not have self-hosted runners)
```

.github/workflows/ubuntu-parallel.yaml
BenjaminRodenberg: Done.

BenjaminRodenberg:
What's still a bit irritating here is that `ubuntu-parallel` actually still hangs. I assume this is due to the policy of github about running CI jobs using workflows from forks. If a fork tries to merge into the main repository it uses the workflow configuration of the main repository. Otherwise outsiders could potentially execute malicious code on internal (self-hosted) runners. (see [here](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository) for details)

To verify the correct behavior of this PR I set up some PRs for testing this on my fork:

* https://github.com/BenjaminRodenberg/elmerfem/pull/2
* https://github.com/BenjaminRodenberg/elmerfem/pull/3

Currently, I'm not sure why some jobs don't get triggered. But I will check this again tomorrow.

@mmuetzel Thanks for the feedback and support so far :wave:

mmuetzel:
> What's still a bit irritating here is that `ubuntu-parallel` actually still hangs.

Are you referring to a longer delay until the job starts? That would probably be because there is only one runner that can pick up jobs from that workflow. So, it might have to clear a queue before it can start the job from this PR.

Or does it actually hang (while building or while testing or during another step)?

BenjaminRodenberg:
> > What's still a bit irritating here is that `ubuntu-parallel` actually still hangs.
> 
> Are you referring to a longer delay until the job starts? That would probably be because there is only one runner that can pick up jobs from that workflow. So, it might have to clear a queue before it can start the job from this PR.
> 
> Or does it actually hang (while building or while testing or during another step)? 

I think the issue was the delay you mentioned. Over night everything in https://github.com/BenjaminRodenberg/elmerfem/pull/2 ran as expected (especially `ubuntu-parallel` is skipped).

So from my perspective this PR is ready to merge.

mmuetzel:
Looks good to me, too. 👍

PR #769, opened 2026-02-05
Projector update

raback:
This includes some cleaning up of some mortar projectors and division of the bloated MeshUtils.F90 into four modules. If somebody has been working with mortars this may pose a challenge for future merges.


juharu:
Looks good, thanks!

mmalinen:
Yes, this looks good

PR #772, opened 2026-02-09
Get coupling grid information from Elmer internals instead of file

BenjaminRodenberg:
Related to https://github.com/ElmerCSC/elmerfem/issues/723 (only SSA)

This is a feature branch contributing to `yac-coupling`.

@ClemensSchannwell tested this code on my fork https://github.com/BenjaminRodenberg/elmerfem/pull/1.


raback (approved):
Looks good to me and only affects the coupling!

Personally I like to inline the messages more so:
WRITE(Message,*) 'Running with ',ParEnv % PEs ,' partitions'
      CALL INFO(SolverName,Message,Level=3)

becomes 
CALL Info(SolverName,'Running with '//I2S(ParEnv % PEs)//' partitions',Level=3)
which also has a spacing more appealing to the eye. 

But this is merely a tip. I2S is used a lot in the code.

PR #773, opened 2026-02-11
Addrint valuelist and _Finalize

juhanikataja:
Added two list subroutines:
- `ListAddAddressInteger`: adds integer of kind `AddrInt` to a valuelist. Uses `valuelistentry_t % PROCEDURE` field for that.
- `ListGetAddressInteger`: gets that integer of kind `AddrInt`.

Added mechanism for calling `<SolverName>_Finalize` subroutine if exists similarly to `<SolverName>_Init/_Init0`. The finalizer is called prior to calling FreeSolver. It is meant for solvers to release resources not managed by Elmer library (e.g. pointers managed by above `ListAddAddressInteger`/`ListGetAddressInteger` routines).


raback:
Seems legit and passes all tests. Nice feature!

PR #777, opened 2026-02-16
Only support YAC built with CMake

BenjaminRodenberg:
CMake support in YAC has been released.

This PR updated the YAC coupling to use CMake features to simplify the building process. Merging this PR will require users to build YAC with CMake. Therefore, this is a breaking change in terms of the `yac-coupling` branch.


cmake/Modules/FindYAC.cmake
```diff
@@ -9,6 +9,13 @@
 #  YAC_FOUND              - True if YAC found including required interfaces (see below)
 #  YAC_LIBRARIES          - All yac related libraries.
 #  YAC_INCLUDE_DIR        - All directories to include.
+#  YAC_VERSION            - Version of YAC found
+
+# Minimum required YAC version (must have CMake config support)
+# Note: Currently a soft requirement with fallback to manual search.
+#       Will become a hard requirement (remove fallback) in the future.
+
+SET(YAC_MINIMUM_VERSION "3.12.0")
```
BenjaminRodenberg: This will be updated to the correct version as soon as YAC supports CMake builds.

CMakeLists.txt
```diff
@@ -637,7 +608,9 @@ IF (WITH_YAC)
     LINK_DIRECTORIES(${YAC_LIBRARIES})
     ADD_DEFINITIONS(-DHAVE_YAC)
   ELSE()
-    MESSAGE(STATUS "  ${YAC_FAILMSG}")
+    MESSAGE(STATUS "YAC not found ")
+    MESSAGE(WARNING "Please be aware that Elmer will only be able to find YAC"
+                    " if YAC has be build with CMake.")
```
BenjaminRodenberg:
```suggestion
                    " if YAC has been build with CMake.")
```

CMakeLists.txt
```diff
@@ -626,6 +594,9 @@ IF (WITH_YAC)
 
   FIND_PACKAGE(YAC)
 
+  GET_TARGET_PROPERTY(YAC_INCLUDE_DIR YAC::yac_core INTERFACE_INCLUDE_DIRECTORIES)
```
BenjaminRodenberg:
```suggestion
  GET_TARGET_PROPERTY(YAC_INCLUDE_DIR YAC::yac_mci INTERFACE_INCLUDE_DIRECTORIES)
```

Not sure if this really matters but from the Elmer perspective we only use yac_mci.

nils3er: Its easy: If you have `USE yac_core` in your code you have to link `yac_core`. If you have `USE yac` in the code you have to link `yac_mci`.

cmake/Modules/FindYAC.cmake
BenjaminRodenberg: Not needed anymore.

BenjaminRodenberg:
This PR is good to go from a technical point of view. However, we still have to wait for the following points:

- [x] YAC release supporting CMake (currently WIP, but planned for one of the next releases)
- [x] ~~ICON release containing this YAC version (since our test case uses ICON and the YAC version bundled with ICON)~~ using `--external-yac` for now

BenjaminRodenberg:
[YAC 3.15.0](https://gitlab.dkrz.de/dkrz-sw/yac/-/tree/v3.15.0) now supports CMake. Still waiting for the ICON release containing this YAC version (or a newer one) before merging this PR. Reason: Merging this PR breaks compatibility with older YAC versions that are currently still bundled with ICON.

BenjaminRodenberg:
I did a test run today with @Clemensschannwell. The CMake build works generally, but we were facing some integration problems with ICON and decided to wait for the upcoming YAC release 3.18.0.

BenjaminRodenberg:
YAC 3.19 contains a fix that was still required for Lapack, see https://gitlab.dkrz.de/YAC/YAC-dev/-/merge_requests/614. So in principle we could do another test run.

BenjaminRodenberg:
Me and @ClemensSchannwell successfully used YAC 3.21.0 (configured and built with CMake) with this setup. ICON uses an external YAC and not the bundled one.

We will require YAC 3.21.0 as minimum version to simplify checking for errors. I will add this update in a moment and merge.

PR #783, opened 2026-02-27
Allow user to configure source coordinate reference system

### LINKED ISSUE #779: YAC coupling: Make projection user-configurable
The projection `"EPSG:3413"` is currently hard-coded:

https://github.com/ElmerCSC/elmerfem/blob/df4bd30697f972184e784e61df6fd51d2d66c1f6/fem/src/project_to_lonlat.c#L42-L44

We need to support arbitrary projections. Therefore, this should be exposed to the user and configurable (via the `.sif` file).

@tzwinger hope this summarizes our discussion and the goal sufficiently? I can take care of this.

BenjaminRodenberg:
Closes #779 

# Todos

- [x] Clarify which projections are needed for our application
- [x] Potentially remove some duplicate code by using functionality from `elmerice/Utils/ProjUtils.F90`


BenjaminRodenberg:
@fgillet would you mind reviewing? I switched the projection to Elmer's internal solution. I hope I'm using everything correctly but I also did some comparison of the old and new method side by side and things look good (my debugging code is still included in [d6334e4](https://github.com/ElmerCSC/elmerfem/pull/783/commits/d6334e4f44d22585ed8cb59b446263df056b4f0d) if you are interested in this intermediate state).

Main points:

* I do not need the external dependency Proj anymore because Elmer provides the functionality. These are a large portion of the deletions. I guess you can ignore most of them.
* I use the functionality from `ProjUtils`. This gave me a linker error first because it is not a part of Elmer Core. So I moved the mesh preparation outside of the core and into `yac2elmer`.

fgillet:
@BenjaminRodenberg looks good for me;
Maybe 2 suggestions:

- The projections are not restricted to EPSG:3413 and 3031; its depends on the two parameters "central_meridian" and "latitude_of_origin"; So maybe if you want to restrict to EPSG: 3413 and 3031 for your coupling you could add a check that the values have been correctly set by the user?
- Maybe you could add a default paremeter in xy2LonLat if you want to get rad instead of deg as the computattion are in rad but transform to deg. afterwards; so it would avoid rad2deg * deg2rad?

some remarks also for futur developments:

- Also you need to be aware for you coupling that there can be some situations in parallel where you need halo elements (Discontinuous galerkin, passive/active for calving) so you may have duplicate nodes and elements as some are shared by the partions; so there is some differences between the bulk and active elements;
- the coupling seems to be restricted to 2D plan view meshes for the moment; There is some elmer utilities than can be used if you want to apply this to the (surface) boundary of a 3D meshes.

BenjaminRodenberg:
Hi @fgillet First of all, thanks for the reply and the review!

Just a heads-up: I did not find the time to work on this in March but it's definitely on my list.

tzwinger:
I am afraid, I laid  havoc on that PR with accepting the PR #800 - @BenjaminRodenberg

BenjaminRodenberg:
@tzwinger well... this PR is already lying around too long. Took care of rebasing this PR onto `yac-coupling` :+1:

BenjaminRodenberg:
@fgillet I worked on your two suggestions above. And added consistency checks in b91ebc5c0092b3d61a5bfe2120c295f4fbe72fef and an optional argument in 06b1ac7390fd9fed98d8c05a3821f0fa2453bb98.

The new argument introduced in 06b1ac7390fd9fed98d8c05a3821f0fa2453bb98 is only applied for "polar stereographic north/south" and ignored for other projections since there was no `deg2rad` in the first place. Hopefully this is correct. But would appreciate your feedback here.

Thanks also for the remarks regarding future development! 3D scenarios are on our radar but we would like to finalize the 2D case first and then continue with 3D.

fgillet:
Thanks for the update; the other projections will return deg. which is the default;

PR #787, opened 2026-03-09
Fix segfault in mortar projectors with discontinuous Galerkin

db434:
I kept hitting segfaults when combining discontinuous Galerkin with radial projectors. Simple test case: [dg-radial.zip](https://github.com/user-attachments/files/25843307/dg-radial.zip).

This fix adds `NotDG=.TRUE.` at a few places so nodes are indexed correctly. Only the final two of these (in `AddProjectorWeak1D`) are needed to fix my example, but my coding agent suggested applying the same fix to `NormalProjectorWeak1D` and `AddProjectorWeakGeneric`.

### DIFF
--- a/fem/src/MortarUtils.F90
+++ b/fem/src/MortarUtils.F90
@@ -2853,7 +2853,7 @@ CONTAINS
         
         Element => BMesh1 % Elements(ind)        
         
-        nd = mGetElementDOFs(Indexes,Element)
+        nd = mGetElementDOFs(Indexes,Element,NotDG=.TRUE.)
         n = Element % TYPE % NumberOfNodes
         ne = 2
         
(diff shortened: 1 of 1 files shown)


raback:
Well detected! Who is your "coding agent". Now it definately fixes the core dump but are you getting continuity over the interface as you would expect. When I changed the inner radius temperature to, say 100, the is a jump between the two bodies. I guess you would like to have a continuous field over the interface?

db434:
My coding agent is Claude.

In my use case, I have a Heat Gap boundary condition at the interface, so I specifically don't want a continuous field. The behaviour looks sensible to me.

Heat Gap Coefficient = 1:
image

Heat Gap Coefficient = 1000:
image

Heat Gap Coefficient = 1000000:
image

raback:
Sorry, pushed a fix directly where I added three more instances of NotDG. Thanx!

PR #790, opened 2026-03-17
Make runoff a cell variable to permit conservative mapping via YAC

ClemensSchannwell:
This PR changes the variable `runoff` from a variable that lives on the vertices of the mesh to a variable that lives on the cell of the mesh. Main reason for this is that it permits conservative interpolation between EBFM and Elmer/Ice so that we conserve mass when passing fluxes.


raback:
Looks good!

PR #799, opened 2026-03-25
nix-workflows: Remove argument that is no longer supported by the action

mmuetzel:
The nixbuild/nix-quick-install-action action no longer(?) supports the parameter `nix_path`.

@mk3z confirmed that it can be removed:
https://github.com/ElmerCSC/elmerfem/pull/795#issuecomment-4118989364


mmuetzel:
The "nix flake check" workflow ran without warnings with that change:
https://github.com/ElmerCSC/elmerfem/actions/runs/23532136877/job/68498276008?pr=799

mk3z (approved):

PR #801, opened 2026-04-11
Elmerice

tzwinger:
Making pull request for back-merger of elmerice - there are several changes from @fgillet on new splitFEM integration rules in SSA solver that should be included upstream


raback:
Looks good to me. Failures on tests are now the same that fail on devel too. Have to work with them this is not the culprint.

PR #802, opened 2026-04-13
Revise macOS compilation instructions for MPI and OpenMP

ZippoLag:
Fixed formatting (using `**`is not being registered as a valid second indentation level). Fixed WITH_OpenMP=FALSE when WITH_MPI=TRUE (setting both to TRUE causes an error)

### DIFF
--- a/compilation_instructions/macOS.md
+++ b/compilation_instructions/macOS.md
@@ -4,23 +4,23 @@ Compilation for macOS
 (note: instructions may be out of date)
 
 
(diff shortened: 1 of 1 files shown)


mmuetzel:
> setting both to TRUE causes an error

Which error does enabling OpenMP and MPI at the same time cause for you on macOS?
It seems to be configuring and building correctly in the CI if both are enabled. E.g.: https://github.com/ElmerCSC/elmerfem/actions/runs/24653943122/job/72082787858#step:5:13

PR #807, opened 2026-05-07
Conforming bc rev

mmalinen:


raback:
Seems ok and test merge does not bring any new issues.

PR #816, opened 2026-06-02
Implement initial ICON-O to Elmer/Ice coupling

BenjaminRodenberg:
Adds coupling of using sea surface temperature and sea surface salinity.

Note: I fetched this branch from @ClemensSchannwell's fork https://github.com/ClemensSchannwell/elmerfem/tree/interface_icon_o to my fork since I could not create a PR from Clemens' fork to the elmer base repository - not sure why...


elmerice/Solvers/yac2elmer.F90
```diff
@@ -98,7 +98,7 @@ SUBROUTINE YAC2Elmer( Model,Solver,dt,TransientSimulation )
   USE elmer_coupling, ONLY: coupling_setup, is_root_rank
   USE elmer_ebfm_coupling, ONLY: elmer_ebfm_interface, t_ice_field, smb_field, &
                                  runoff_field, surface_height_field
-  ! USE elmer_icon_coupling, ONLY: elmer_icon_interface, clt_field, pr_field
+  USE elmer_icon_coupling, ONLY: elmer_icon_interface, t_oce_field, sal_oce_field
```
BenjaminRodenberg: General point about the updated implementation: Previously we had a draft of coupling ICON-A to Elmer in terms of comments. Here, you deleted the comments ( = dead code) and replaced them with actual code for coupling ICON-O.

Generally replacing dead code with proper code is fine, but I would not delete the draft for the coupling to ICON-A if we will need this in the future. We could do the following:

```
USE elmer_icon_coupling, ONLY: elmer_icon_interface  ! general stuff
! USE elmer_icon_coupling, ONLY: clt_field, pr_field  ! ICON-A
USE elmer_icon_coupling, ONLY: t_oce_field, sal_oce_field  ! ICON-O
```

We will potentially merge these `USE ... ONLY ...` statements in the future when the final version is ready but for now I think this is a good pattern for prototyping.

elmerice/Solvers/yac2elmer.F90
```diff
@@ -117,14 +117,14 @@ SUBROUTINE YAC2Elmer( Model,Solver,dt,TransientSimulation )
   CHARACTER(LEN=1024) ::  config_file, model_tstep, coupling_timestep, grid_crs, proj_type
   INTEGER :: i, t, ierr, dt_hours, coupling_hours
   INTEGER, POINTER :: t_icePerm(:), smbPerm(:), runoffPerm(:)
+  INTEGER, POINTER :: t_ocePerm(:), sal_ocePerm(:)
   REAL(KIND=dp) :: central_meridian, latitude_of_origin
   REAL(KIND=dp) :: expected_central_meridian, expected_latitude_of_origin
   CHARACTER(LEN=16) :: expected_central_meridian_str, expected_latitude_of_origin_str
-  ! INTEGER, POINTER :: cltPerm(:), prPerm(:)  ! ICON is not supported at the moment
```
BenjaminRodenberg, line 126: See comment above. There are many more similar code snippets below that I will ignore from here on.

elmerice/Solvers/yac2elmer.F90
```diff
@@ -374,7 +362,20 @@ END SUBROUTINE collect_coupling_grid_data
...
+      CALL INFO(SolverName, 'BEFORE ELMER ICON-O INTERFACE', Level=3)
+      ! couple with ICON-O
+      CALL elmer_icon_interface(is_root_rank)
+      CALL INFO(SolverName, 'AFTER ELMER ICON-O INTERFACE', Level=3)
+      t_oceVar => VariableGet( Mesh % Variables, 'temp_oce' )
+      sal_oceVar => VariableGet( Mesh % Variables, 'sal_oce' )
+      IF ((.NOT.ASSOCIATED(t_oceVar)) .OR. (.NOT.ASSOCIATED(sal_oceVar))) THEN
+        CALL FATAL(SolverName,'Elmer variables not associated')
+      END IF
+       !write over values for nodes
```
BenjaminRodenberg:
```suggestion
      ! write over values for nodes
```

elmerice/Solvers/yac2elmer.F90
BenjaminRodenberg:
```suggestion
! TODO: stub implementation for ICON-A coupling
! CALL elmer_icon_interface(is_root_rank)
! cltVar => VariableGet( Mesh % Variables, 'tas' )
! prVar => VariableGet( Mesh % Variables, 'pr_snow' )
```

What about this commented out code? Maybe make it more explicit that ICON-A coupling is missing at the moment here.

fem/src/elmer_coupling.F90
```diff
@@ -527,7 +527,7 @@ SUBROUTINE print_field_info(elmer_comp_name, elmer_grid_name, field_name)
             elmer_comp_name, elmer_grid_name, field_name) == &
             YAC_EXCHANGE_TYPE_TARGET) THEN
 
-#if YAC_VERSION_GREATER_EQUAL(3, 6, 0)
+#if (YAC_VERSION_MAJOR >= 3) && (YAC_VERSION_MINOR >= 6)
```
BenjaminRodenberg: Why this change? IIRC `YAC_VERSION_GREATER_EQUAL` implements pretty much what you are doing here. I would rather use the implementation `YAC_VERSION_GREATER_EQUAL` provided by YAC instead of reinventing the wheel here.

fem/src/elmer_coupling.F90
```diff
@@ -539,11 +539,11 @@ SUBROUTINE print_field_info(elmer_comp_name, elmer_grid_name, field_name)
 
         IF (yac_fget_field_role( &
               elmer_comp_name, elmer_grid_name, field_name) == &
-              YAC_EXCHANGE_TYPE_SOURCE) THEN
+              YAC_EXCHANGE_TYPE_TARGET) THEN
```
BenjaminRodenberg: Was this buggy/misleading before of why this change? This whole block is a debugging statement so it should not affect the results or lead to a crash if something is wrong here but since you touched it I assume it was somewhat useful for your development - so we should make sure the output is generally helpful. We can also print for both cases (Elmer fields being source and target). Maybe that's a bit verbose if we duplicate the code but I don't think this is a huge problem.

BenjaminRodenberg (commented):
@ClemensSchannwell I can take care of the integration. Could you maybe clarify on the following points to give me a rough direction?

fem/src/elmer_coupling.F90
BenjaminRodenberg: A longer note on this macro: Older YAC versions do not offer `YAC_VERSION_GREATER_EQUAL` but we introduced it in YAC 3.9.3 (see [here](https://yac.gitlab-pages.dkrz.de/YAC-dev/d5/d6f/changelog.html#version_393)). It should be available when you import `#include "yac.h"`.

For older versions this does not help. However, I'm planning to generally only support YAC versions that support building with CMake because this makes building just so much easier in the Elmer ecosystem. We are still waiting for the next ICON release because CMake support has just been released in YAC 3.15.0 and the latest ICON currently bundles YAC 3.14.0_p1.

Long story short: For CMake we still have to wait but for using `YAC_VERSION_GREATER_EQUAL` there is not really a good reason for a (backwards compatible) work-around since you should ideally use the latest ICON and this comes with a YAC providing `YAC_VERSION_GREATER_EQUAL`.

elmerice/Solvers/yac2elmer.F90
ClemensSchannwell: I agree that we should keep it. Even though it is unlikely we will talk directly between Elmer and ICON-A, it is still useful, because ICON-Land lives on the same grid as ICON-A. Can you make sure that I did not screw up and this is all done according to best practises?

elmerice/Solvers/yac2elmer.F90
ClemensSchannwell: As I said above, yes, we should keep all of this. You are absolutely correct.

fem/src/elmer_coupling.F90
ClemensSchannwell: Sorry, this must have slipped in from maybe one of the older code versions. Please revert this change!

fem/src/elmer_coupling.F90
ClemensSchannwell: No, I think this is still a mistake from when I was implementing this. I had some issue that YAC was not getting the correct coupling timestep and changed this. It had obviously nothing to do with the underlying problem and can be reverted.

fem/src/elmer_coupling.F90
```diff
@@ -449,48 +449,46 @@ MODULE elmer_icon_coupling
...
-    nbr_cells = yac_fget_points_size(cell_point_id)
 
-    ! register total cloud cover field in YAC
-    CALL yac_fdef_field( &
-      clt_field_name, comp_id, (/corner_point_id/), 1, clt_collection_size, &
-      timestepstring, YAC_TIME_UNIT_HOUR, clt_field_id);
+    ! register ocean temperature field in YAC
+    CALL yac_fdef_field_mask( &
+      t_oce_field_name, comp_id, (/corner_point_id/), (/boundary_corner_mask_id/), 1, &
+      t_oce_collection_size, timestepstring, YAC_TIME_UNIT_HOUR, t_oce_field_id)
```
BenjaminRodenberg: @ClemensSchannwell if you want to revert to the unmasked field from before: Simply replace `yac_fdef_field_mask` with `yac_fdef_field` and remove the additional argument `(/boundary_corner_mask_id/)`. I believe you can keep the mask around for now until we know what we do with it.

fem/src/elmer_coupling.F90
```diff
@@ -449,48 +455,148 @@ MODULE elmer_icon_coupling
...
 
-  INTEGER :: clt_field_id = -1
-  CHARACTER(LEN=*), PARAMETER :: clt_field_name = "tas"
-  INTEGER :: clt_collection_size = 1
-  DOUBLE PRECISION, PUBLIC, ALLOCATABLE :: clt_field(:,:)
-
-  INTEGER :: pr_field_id = -1
-  CHARACTER(LEN=*), PARAMETER :: pr_field_name = "pr_snow"
-  INTEGER :: pr_collection_size = 1
-  DOUBLE PRECISION, PUBLIC, ALLOCATABLE :: pr_field(:,:)
```
BenjaminRodenberg, line 460: As far as I understood, entirely removing `clt_field` and `pr_field` is intentional and we will not need it anymore.

BenjaminRodenberg (commented):
@ClemensSchannwell can you do a final test of this branch? From my perspective ready to merge.

tzwinger:
Should we merge that one?

elmerice/Solvers/yac2elmer.F90
```diff
@@ -476,11 +538,20 @@ END SUBROUTINE collect_coupling_grid_data
...
+      t_oceVar => VariableGet( Mesh % Variables, 'temp_oce_post' )
+      sal_oceVar => VariableGet( Mesh % Variables, 'sal_oce_post' )
+      IF ((.NOT.ASSOCIATED(t_oceVar)) .OR. (.NOT.ASSOCIATED(sal_oceVar))) THEN
+        CALL FATAL(SolverName,'Elmer variables not associated')
+      END IF
+      ! write over values for nodes
+      DO i=1, Mesh % NumberOfNodes
+        t_oceVar % Values(t_oceVar % Perm(i)) = t_oce_post_field(i,1)
+        sal_oceVar % Values(sal_oceVar % Perm(i)) = sal_oce_post_field(i,1)
+      END DO
```
BenjaminRodenberg, line 553: As mentioned in https://github.com/BenjaminRodenberg/elmerfem/pull/6#discussion_r3508344762 exposing `_pre`/`_post` to the outside world is a bit odd because both fields are implementation details of `yac2elmer`/`elmer_coupling.F90`.

This is just a reminder and I need to look into this.

elmerice/Tests/SubShelfMeltLinearTF/InitOceanVarsSolver.F90
```diff
@@ -0,0 +1,52 @@
...
+  IF (.NOT. Found) CALL FATAL('InitOceanVarsSolver', 'temp_oce Value not found')
+  S_val  = GetConstReal(SolverParams, 'sal_oce Value', Found)
+  IF (.NOT. Found) CALL FATAL('InitOceanVarsSolver', 'sal_oce Value not found')
+
+  Bedrock_val = GetConstReal(SolverParams, 'bedrock Value', BedrockFound)
+
+  Zb_var    => VariableGet(Solver % Mesh % Variables, 'Zb')
+  GM_var    => VariableGet(Solver % Mesh % Variables, 'GroundedMask')
+  Toce_var  => VariableGet(Solver % Mesh % Variables, 'temp_oce')
+  Soce_var  => VariableGet(Solver % Mesh % Variables, 'sal_oce')
```
BenjaminRodenberg, line 35: I removed the `_post` here and used the names `temp_oce` and `sal_oce` as identifiers.

elmerice/Solvers/yac2elmer.F90
```diff
@@ -476,11 +537,20 @@ END SUBROUTINE collect_coupling_grid_data
...
-      ! TODO: stub implementation for ICON coupling
-      ! CALL elmer_icon_interface(is_root_rank)
-      ! cltVar => VariableGet( Mesh % Variables, 'tas' )
-      ! prVar => VariableGet( Mesh % Variables, 'pr_snow' )
+      CALL INFO(SolverName, 'BEFORE ELMER ICON-O INTERFACE', Level=3)
+      ! couple with ICON-O
+      CALL elmer_icon_interface(is_root_rank)
+      CALL INFO(SolverName, 'AFTER ELMER ICON-O INTERFACE', Level=3)
+      t_oceVar => VariableGet( Mesh % Variables, 'temp_oce' )
+      sal_oceVar => VariableGet( Mesh % Variables, 'sal_oce' )
```
BenjaminRodenberg, line 545: These fields are now simply identified via `temp_oce` and `sal_oce` (without `_post`); the `_post` part is now an internal detail of `yac2elmer.F90` and `elmer_coupling.F90`.

One think to keep in mind: We use the labels `temp_oce` and `sal_oce` for two different things now

1. As names for the fields that are sent from ICON to Elmer
2. As internal identifiers inside Elmer

Between 1 and 2 we perform the whole creep mapping. Not sure if this double meaning of the labels is confusing or actually quite elegant. We could, of course also pick a different name for the fields inside elmer if you want.

BenjaminRodenberg (commented):
@ClemensSchannwell I merged your cleanup PR and revised the labels for the fields to hide `_pre`/`_post` better. See b6ecee5f5c468461b4bf19aff9cd61b19e4a4dd8.

Would you mind doing a final test & review?

From my perspective ready to merge.

elmerice/Solvers/yac2elmer.F90
ClemensSchannwell: I think this is fine and I personally like this much better than this whole `_post` business. I would argue that this is just encapsulation and really the user is not supposed to work with these intermediate fields anyway. 

I have tested this branch and everything works as expected. So from my point, it can be merged.
Thanks for sorting this out!

PR #821, opened 2026-06-09
Drop incorrect quoting in compiler wrappers.

mmuetzel:
This basically reverts the changes from 7f3b3c7194d09abf40bb5e2aa608bb7021cb69bd.

Quote the string that should be copyable and run as a shell command instead.


juharu:
Thanks Markus!

PR #822, opened 2026-06-10
More compiler wrapper changes.

mmuetzel:
Remove `elmerf90-nosh` because it is no longer needed. (See https://github.com/ElmerCSC/elmerfem/issues/814#issuecomment-4666901859)

Reduce code duplication between the two compiler wrappers `elmerf90` and `elmerld` by moving shared parts to a header file.


juharu:
Thanks!

Issue #195, opened 2020-01-27
mpistub not found error on installation of release 8.4


ketancmaheshwari:
cmake command is: cmake -DWITH_ELMERGUI:BOOL=FALSE -DWITH_MPI:BOOL=FALSE -DCMAKE_INSTALL_PREFIX=/software/dev_tools/swtree/cs400/elmer/8.4/centos7.5_gnu5.3.0 ..

CMake Error at fem/src/CMakeLists.txt:135 (FILE):
  FILE COPY cannot find
  "/software/dev_tools/swtree/cs400/elmer/8.4/centos7.5_gnu5.3.0/elmer-8.4/fem/src/mpif_stub.h".

CMake Error at fem/src/CMakeLists.txt:137 (FILE):
  FILE RENAME failed to rename

    /software/dev_tools/swtree/cs400/elmer/8.4/centos7.5_gnu5.3.0/elmer-8.4/fem/src/mpif_stub.h

  to

    /software/dev_tools/swtree/cs400/elmer/8.4/centos7.5_gnu5.3.0/elmer-8.4/fem/src/mpif.h

  because: No such file or directory

Any suggestions to fix this?

PR #824, opened 2026-06-11
Use CMake generator expressions to get file names of ElmerIce libs for `elmer{f90,ld}`

mmuetzel:
Linking to a library with `-l` requires that its file name begins with `lib` on Linux (and macOS?). So, the current implementation did not work for the ElmerIce libraries.

~~However, it shouldn't be necessary to link to the dependencies of the (shared) libelmersolver library anyway. (That would only be necessary if it were a static library.) So, instead of trying to fix that, remove linking to these (optional) dependencies entirely.
If a user requires to link to these libraries, they can pass the respective flags explicitly.~~
(See: https://github.com/ElmerCSC/elmerfem/issues/814#issuecomment-4678252655)

Edit: 
Instead use `-l:` with the full name of the libraries. Use CMake generator expressions the get the file names from the respective CMake targets.


mjfwest:
Think this just broke it, with 
```
12:49:21   #195 13.94 CMake Error at fem/src/CMakeLists.txt:126 (FILE):
12:49:21   #195 13.94   Error evaluating generator expression:
12:49:21   #195 13.94
12:49:21   #195 13.94     $<TARGET_FILE_NAME:ElmerIceSolvers>
12:49:21   #195 13.94
12:49:21   #195 13.94   No target "ElmerIceSolvers"
12:49:21   #195 13.94
12:49:21   #195 13.94
12:49:21   #195 13.94 CMake Error at fem/src/CMakeLists.txt:126 (FILE):
12:49:21   #195 13.94   Error evaluating generator expression:
12:49:21   #195 13.94
12:49:21   #195 13.94     $<TARGET_FILE_NAME:ElmerIceSolvers>
12:49:21   #195 13.94
12:49:21   #195 13.94   No target "ElmerIceSolvers"
12:49:21   #195 13.94
12:49:21   #195 13.94
12:49:21   #195 13.94 CMake Error at fem/src/CMakeLists.txt:126 (FILE):
12:49:21   #195 13.94   Error evaluating generator expression:
12:49:21   #195 13.94
12:49:21   #195 13.94     $<TARGET_FILE_NAME:ElmerIceSolvers>
12:49:21   #195 13.94
12:49:21   #195 13.94   No target "ElmerIceSolvers"
12:49:21   #195 13.94
12:49:21   #195 13.94
12:49:23   #195 16.05 -- Generating done```

mmuetzel:
Which version of CMake was that with?

I now realized that `TARGET_EXISTS` was added in CMake 3.12. The minimum CMake version is currently 3.10. Are you using an older version than 3.12?
Which distribution are you using?

PR #826, opened 2026-06-13
Bump minimum required CMake version to 3.12

mmuetzel:
That version is when `TARGET_EXISTS` was first introduced. https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:TARGET_EXISTS

According to AI, most popular distributions should have a newer version than that by now:

| Distribution                    | CMake version                     |
| ------------------------------- | --------------------------------- |
| Ubuntu 24.04 LTS                | 3.28                              |
| Ubuntu 22.04 LTS                | 3.22                              |
| Ubuntu 20.04 LTS (end-of-life)  | 3.16                              |
| Debian 12 (Bookworm)            | 3.25                              |
| Fedora 40 / 41 /42              | 3.29 to 3.31                      |
| RHEL 9 / Rocky Linux 9          | 3.20 to 3.26                      |
| RHEL 8                          | 3.20 to 3.26                      |
| SUSE Linux Enterprise Server 15 | 3.20                              |
| Arch Linux                      | Latest upstream (rolling release) |


mmuetzel:
Fwiw, CMake 3.12 was released on 17 July 2018 (so almost 8 years ago).
https://www.kitware.com/cmake-3-12-0-available-for-download/

mmuetzel:
This change might no longer be necessary if #827 is merged.

PR #832, opened 2026-06-30
ElmerGrid, add reading of Gmsh Format 2.2 binary mesh files.

richb2k:
Synchronize with ElmerGUI.

### DIFF
--- a/elmergrid/src/egconvert.c
+++ b/elmergrid/src/egconvert.c
@@ -3884,7 +3884,7 @@ static int LoadGmshInput1(struct FemType *data,struct BoundaryType *bound,
   int noknots = 0,noelements = 0,maxnodes,dim;
   int elemind[MAXNODESD2],elementtype;
   int i,j,k,allocated,*revindx=NULL,maxindx;
-  int elemno, gmshtype, regphys, regelem, elemnodes,maxelemtype;
+  int elemno, gmshtype, regphys, regelem, elemnodes=0,maxelemtype;
   FILE *in;
   char *cp,line[MAXLINESIZE];
 
@@ -4044,7 +4044,7 @@ static int LoadGmshInput2(struct FemType *data,struct BoundaryType *bound,
   int noknots = 0,noelements = 0,nophysical = 0,maxnodes,dim,notags,nodenum;
   int elemind[MAXNODESD2],elementtype;
   int i,j,k,allocated,*revindx=NULL,maxindx;
-  int elemno, gmshtype, tagphys=0, taggeom=0, tagpart, elemnodes,maxelemtype;
+  int elemno, gmshtype, tagphys=0, taggeom=0, tagpart, elemnodes=0,maxelemtype;
   int tagmat,verno,frcount;
   int physvolexist, physsurfexist;
   FILE *in;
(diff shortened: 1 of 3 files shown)


elmergrid/src/egconvert.c
```diff
@@ -4118,9 +4136,69 @@ static int LoadGmshInput2(struct FemType *data,struct BoundaryType *bound,
       cp = line;
       noelements = next_int(&cp);
 
+      int inElems[126];
+      // inElems size is based on gmsh element type 93, 125-node fourth order hexahedron
+      // if newer and larger elements are added, then increase nodeTag size to match
+      if(elemnodes > 125) {
+        printf("error: number of nodes per element of %d, exceeds 125\n",elemnodes);
```
mmuetzel, line 4143: It looks like `elemnodes` is used unintialized. 
Afaics, it's first assigned a value further down in the function (new line number 4164). So its value here is undefined.

richb2k: Okay, a fix has been added.  Thanks for pointing it out.

mmuetzel: Nice!
It looks like that fixed the test errors on macOS. 👍

raback: Thanx for the fixes! One day we will be bug free...

mmuetzel: Not to spoil anything. But this was "just" a fix to a new bug in the first iteration of this PR. It doesn't fix a bug that existed before this PR. (I haven't checked whether there were other code paths that potentially tried to use uninitialized values. So, maybe it fixes existing undefined behavior there.)

In any case, it adds a new feature to ElmerGrid. 👍

mmuetzel: Come to think of it. GitHub has [CodeQL](https://codeql.github.com/) - a static analyzer that might detect patterns like that.
It currently doesn't have support for Fortran though. So, it would be useless for the vast majority of the source code. But it could potentially help with the parts that are written in C or C++.
Would it make sense to add a CodeQL workflow for that?

richb2k: Perhaps it would help to ask the compiler to check for more warnings.

Currently, compiling just ElmerGrid with no C Flags, produces 5 warnings.
Compiling ElmerGrid with -Wmaybe-uninitialized produces a total of 29 warnings.
Compiling ElmerGrid with -Wall -Wextra produces 522 warnings.

mmuetzel: Good idea!

While the "maybe-uninitialzed" warnings could flag scenarios which can't actually occur on runtime, it is probably still a good idea to address them where possible.

Would you like to open a PR (or more than one if that makes sense) to address these issues?

Building ElmerFEM produces a huge amount of warnings for me. That makes it difficult to find what is most important.
For a test, it could potentially make sense to build with `-Werror=maybe-uninitialized`. 
Maybe, even add that to the CI configuration. (However, that is only possible if none of these warnings are emitted in upstream headers that Elmer includes in some files.)

richb2k: I'll work on reducing the number of warnings in ElmerGrid.  It may take a while, so no promises for a quick fix.

PR #842, opened 2026-07-16
Create release assets that include Git submodules.

### LINKED ISSUE #780: Zoltan sub-repo missing in release tarball
The source tarball that can be downloaded as one of the assets in https://github.com/ElmerCSC/elmerfem/releases/tag/release-26.1 doesn't contain data from the Zoltan sub-repository. The folder `contrib\Zoltan_v3.83` in that source tarball is empty.

Trying to configure with `-DWITH_Zoltan=ON` using that tarball leads to the following error:
```
-- Building Zoltan mesh repartitioner
CMake Error at CMakeLists.txt:343 (ADD_SUBDIRECTORY):
  The source directory

    D:/repo/MSYS2/MINGW-packages/mingw-w64-elmerfem/src/elmerfem-release-26.1/contrib/Zoltan_v3.83

  does not contain a CMakeLists.txt file.


--   Zoltan library:
-- ------------------------------------------------
-- ------------------------------------------------
```

mmuetzel:
GitHub automatically creates release assets using `git archive`. That does not include Git submodules (i.e., no Zoltan).

Add a workflow that creates assets that include Git submodules.

The workflow can be triggered manually to create assets for arbitrary refs. In that case, these assets are optionally uploaded to existing releases. The latter only works if the ref is a tag though.

The workflow also triggers automatically when a release is being published in which case it uploads the created assets to the new release.

This fixes #780.

To be able to upload the assets to the release, someone with sufficient privileges would need to create a token for that and add it as the secret `GITHUB_RELEASE_TOKEN` in this repository.
@raback: Would you be interested in that and be able to do that?

I couldn't test whether uploading the release assets actually works. But I hope it does.
At least in the created workflow artifacts, the Zoltan submodule is included in my testing.

### DIFF
--- a/.github/workflows/create-release-assets.yaml
+++ b/.github/workflows/create-release-assets.yaml
@@ -1,0 +1,84 @@
+name: create-release-assets
+# GitHub automatically creates release assets using `git archive`.
+# That does not include submodules (i.e., no Zoltan).
+# This workflow creates release assets including submodules.
+
+on:
+  # Manually trigger workflow to upload assets to existing release or create
+  # artifacts for arbitrary refs.
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: "Tag, branch, or SHA to archive"
+        required: true
+        default: "devel"
+      publish_release:
+        description: "Upload archives to release"
+        type: boolean
+        default: false
+  # Automatically trigger workflow when publishing a new release
+  release:
+    types: [published]
+
+jobs:
+  create-archives:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository with submodules
+        uses: actions/checkout@v7
+        with:
+          submodules: recursive
+          fetch-depth: 0
+
+      - name: Install git-archive-all
+        run: pip install git-archive-all
+
+      - name: Create archives including submodules
+        run: |
+          mkdir -p dist
+          echo Creating gz-compressed tarball...
+          git-archive-all dist/elmerfem-${{ github.event.inputs.ref || github.event.release.tag_name }}.tar.gz
+          echo Creating zip archive...
+          git-archive-all dist/elmerfem-${{ github.event.inputs.ref || github.event.release.tag_name }}.zip
+
+      # Upload as workflow artifacts (if triggered manually)
+      - name: Upload .tar.gz as workflow artifact
+        if: github.event_name == 'workflow_dispatch'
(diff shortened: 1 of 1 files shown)


mmuetzel:
@raback: Looking at the release dates of 26.1 and 26.2, I guess that 26.3 will probably be tagged in a week or so.

Could you please consider merging this before the next release?

If you'd like some instructions on how to create and set-up a token for this, I can try to put something together.

raback:
Hi Markus. I added a "GITHUB_RELEASE_TOKEN" - not quite share what it does. Does it mean that next time a release is created .zip and .tar.gz will be created automatically? That's nice. Thanx!

mmuetzel:
They should be created automatically if all works well. (At least, that was the plan.)

You could check parts of that workflow by attempting to create release assets for a previous release. E.g., trigger a workflow passing "release-26.2" as the reference and ticking "Upload archives to releases":
image
(Different to that screenshot from my fork, you'll probably need to select the `devel` branch if that isn't selected automatically.)

That should be creating release tarballs from that tag including the Zoltan submodule and add them to that release if everything is set up correctly.

Does that work?

For future releases, you shouldn't need to trigger the workflow manually. It should be triggering automatically by itself.

raback:
It seems to work at least partially. However, for tar.gz there was an issue:
https://github.com/ElmerCSC/elmerfem/actions/runs/30261319203/job/89961626321

Node 20 is being deprecated. This workflow is running with Node 24 by default. If you need to temporarily use Node 20, you can set the ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true environment variable.

Also, it seems that this broke that msys runners. For me it seems that I should add a packboy package "nlohmann-json:p" in build-windows-mingw.yaml, whatever this does...

mmuetzel:
Oops.

I didn't think of the fact that `github.event.release.upload_url` won't be set if the workflow didn't trigger on a release event. That case is hopefully covered by #851.

> Also, it seems that this broke that msys runners. For me it seems that I should add a packboy package "nlohmann-json:p" in build-windows-mingw.yaml, whatever this does...

I also noticed that in the CI run of #849.
It's is very likely unrelated to the change here. But it is probably a new indirect build dependency (caused by a package update in MSYS2).
You are right, a new `nlohmann-json:p` for the `pacboy` variable should fix that. I came to the same conclusion and opened #850 with that change.

PR #843, opened 2026-07-16
Overhaul generation of CPack bundles for Windows

### LINKED ISSUE #834: Windows installers with mpi not building.
The most recent Windows installer, with mpi and with gui, last successfully built on 6 June 2026. The non-mpi versions, with and without gui successfully built yesterday.

mmuetzel:
The current packaging rules include a lot of manual lists of files and apparently require (manually?) creating a folder layout in which files need to be copied.
At least to me, it is not clear immediately where which files needs to be copied to when attempting to create a Windows installer with CPack.

This series of commits attempts to automate as much of that process as possible by using a tool to collect dependent DLLs from MSYS2.
I haven't (yet?) looked into `BUNDLE_STRIPPED_GFORTRAN` and how it might be generalized. I'm unsure what would be expected in the directory `${CMAKE_CURRENT_SOURCE_DIR}/../stripped_gfortran` if that flag is enabled.

Almost everything is already built in the existing Windows workflow. So, I figured it wouldn't hurt to adapt it so that it creates these CPack bundles and uploads them as artifacts as part of the action.

I *guess* these changes would still work with wherever the Windows nightlies are built currently. But I'm not sure.
@raback: Could you please check that when you come around to it?

These changes could potentially help resolve #834. (Although, I don't know what is actually failing there.)


mmuetzel:
The last commit in this PR (the one that skips `libarpack.dll` and `libparpack.dll` if they are built as part of ElmerFEM) might need to be reverted if #844 should be accepted.

mmuetzel:
It looks more likely to me that the bundled ARPACK/PARPACK will be upgraded to a current version of ARPACK-ng (without switching to static libraries).
So, the PR is probably ready as it is (and my last comment can be disregarded).

That is if the new approach is compatible to how the nightly installers are created. Alternatively (or additionally?), users could be pointed to the workflow artifacts in the CI here if they'd like to test the newest shiniest version of Elmer on Windows.

raback:
I checked our archaic Windows builder and it has no problems with this PR. Hope it helps generation of such installers in the futurel. Thanx!

mmuetzel:
Thanks for merging.

With this being merged, the installers will be uploaded as artifacts for each run of the MinGW workflow.
They are downloadable, e.g., at the bottom of this page for that run: https://github.com/ElmerCSC/elmerfem/actions/runs/30269754852

I'm not sure how long these assets are available. But I think it is possible to change the "retention days" (or something along those lines) somewhere in the repository settings if you like.
By default, I think they'll disappear after some time.

PR #849, opened 2026-07-27
Allocate buffer for BSend in AMGXSolver

dancaseb:
Similar to ROCSolver, attach enough buffer space before calling MPI_BSend in the AMGXSolver (when setting up the global matrix and ranks communicate with eachother regarding owned and non-owned rows).


raback:
Seems legit! Thanx!

PR #851, opened 2026-07-27
Overhaul workflow that creates release assets

mmuetzel:
If the workflow is triggered manually (i.e., on a `workflow_dispatch` event instead of a `release` event), `github.event.release.upload_url` is empty.

In that case, the upload URL must be determined in a separate step.

### DIFF
--- a/.github/workflows/create-release-assets.yaml
+++ b/.github/workflows/create-release-assets.yaml
@@ -59,26 +59,54 @@ jobs:
           path: dist/elmerfem-${{ github.event.inputs.ref || github.event.release.tag_name }}.zip
           archive: false
 
+      - name: Determine upload_url
+        if: github.event_name == 'release' || github.event.inputs.publish_release == true
+        id: determine_upload_url
+        env:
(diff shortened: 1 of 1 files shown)


mmuetzel:
I added another commit that switches to using `softprops/action-gh-release` action instead of `actions/upload-release-asset` (because the latter is no longer maintained).

Additionally to being the recommended and maintained replacement for the previously used action, this allows simplifying the workflow a bit.
In particular, it should no longer be necessary to use a personal access token (PAT). Instead, the built-in `GITHUB_TOKEN` should be just fine.

@raback: If it turns out that this is working, I'd recommend deleting the PAT again (to avoid any potential issues in case that token ever leaks).

mmuetzel:
Thanks for merging.

Does it now succeed if you trigger the workflow for an existing release?

raback:
Yes it does! Nice feature.

mmuetzel:
Thanks for testing.
It should be uploading these assets to the release automatically. (At least that was the plan.)
#852 will hopefully fix that.

PR #852, opened 2026-07-28
Fix some issues with uploading assets to a release

mmuetzel:
I didn't think about the option to tag a release in my fork of this repository to be able to test the new workflow there first.
Now that I did that, I found a few more issues with the current workflow.

With the proposed changes, I was able to have the workflow upload the created tarballs to a test release that I created before:
https://github.com/mmuetzel/elmerfem/releases/tag/test-release-1
The corresponding workflow run:
https://github.com/mmuetzel/elmerfem/actions/runs/30338385559

I then created a new test release:
https://github.com/mmuetzel/elmerfem/releases/tag/test-release-2
Publishing triggered the workflow automatically:
https://github.com/mmuetzel/elmerfem/actions/runs/30339172629
And the created tarballs appeared as assets for that release when the run completed.

I hope that will also work in this repository.

### DIFF
--- a/.github/workflows/create-release-assets.yaml
+++ b/.github/workflows/create-release-assets.yaml
@@ -67,9 +67,11 @@ jobs:
           path: dist/elmerfem-${{ github.event.inputs.ref || github.event.release.tag_name }}.zip
           archive: false
 
-      # Fail early if workflow_dispatch requests publishing but no release exists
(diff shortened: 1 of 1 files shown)


mmuetzel:
Thanks for merging.

Could you please re-run the workflow for an existing release to make sure that it will append the assets correctly now?

PR #853, opened 2026-07-29
Check out the repository at the given reference for tarball creation

mmuetzel:
I forgot to include the reference in the step that checks out the repository. That means that the action always created tarballs for the current head of the default branch. They just had a different name.

To fix that, actually use the given reference when checking out the repository.

I'm sorry for that oversight.

### DIFF
--- a/.github/workflows/create-release-assets.yaml
+++ b/.github/workflows/create-release-assets.yaml
@@ -35,6 +35,7 @@ jobs:
       - name: Checkout repository with submodules
         uses: actions/checkout@v7
         with:
+          ref: ${{ github.event.inputs.ref || github.event.release.tag_name }}
           submodules: recursive
           fetch-depth: 0
 


mmuetzel:
@raback: Could you please have a look at this shortly?
This means that the assets that this action uploaded before didn't actually match the tagged release. That means that users might be downloading the wrong tarballs currently.
Fortunately, running the action again just overrides the incorrect release assets.
Could you please merge this, and trigger the workflow for the existing releases again?

mmuetzel:
To clarify: The release assets appearing, e.g., at the bottom of this page are currently wrong:
https://github.com/ElmerCSC/elmerfem/releases/tag/release-26.2

In my humble opinion, it would be important to merge this PR and trigger the workflow for the three past releases again as soon as possible.

raback:
Thanx Markus! I run the workflow again (for 26 series). It seems though that the files with old naming convention "Source code*" are kept while the new ones are "elmerfem-release*".

mmuetzel:
Thanks!

As far as I can tell, the "Source code" assets are created automatically by GitHub when creating a release (using `git archive` in the background).
I don't know if that can be suppressed or if they can be deleted manually.
I can try to read up on this if you like them being removed.

raback:
Ok, my bad. I misunderstood. I think it is fine as it is. Thank you again. You always find new ways to improve things.

PR #854, opened 2026-07-30
Support building ElmerGrid with an external Metis

mmuetzel:
Add `EXTERNAL_METIS` to select whether to build ElmerGrid with an external or the bundled Metis.

This new flag serves a similar purpose as `EXTERNAL_UMFPACK`, `EXTERNAL_ARPACK`, or `EXTERNAL_PARPACK`.

Currently, MSYS2 is applying this patch on to of the release tarball of ElmerFEM:
https://github.com/msys2/MINGW-packages/blob/0fd872c997a7469908ff7aee7cce83b951e2d59d/mingw-w64-elmerfem/001-use-system-metis.patch
(I'm not sure why that matters to them though.)

Adding this flag would allow dropping that patch from their packaging rules.

This change also adapts the CI rules for MSYS2 to make sure that using their Metis library works correctly.


raback:
Seems to run now succesfully on all platforms. I guess Metis has been really steady but still it is nice that we can eliminate the internal dependence.

PR #855, opened 2026-07-31
Fail AMGXSolver if the solver doesn't converge

dancaseb:
User can toggle if the program aborts completely or if it prints out the solution using   Linear System Abort Not  Converged flag


raback:
Looks good!

PR #856, opened 2026-07-31
SaveScalars: Do not assume frequency unit

marioalexis84:
Do not assume that the unit of frequency is Hertz


raback:
Sorry, I am of slightly different opinion. It is important to make a distinction between units of Hz and unit of rad/s. I know that the time unit need not to be second but because we want to confuse the rad/[t] interpretation I would rather let the unit stay. Most people are happy with seconds. Even US uses that as base unit. And the rare physicist who use nanoseconds, Plank time etc. can handle the issues...

PR #862, opened 2026-08-07
Spawn and wait for new process in compiler and linker wrappers on Windows

mmuetzel:
Apparently, exit codes of replaced processes are not correctly returned on Windows. (See https://github.com/ElmerCSC/elmerfem/issues/858#issuecomment-5215231305)

Instead of replacing a process with `exec`, `spawn` a new process, wait for its completion and return its exit code from the wrapper process.


mmuetzel:
Oops. Only tested on Windows.

Hopefully better now.

PR #865, opened 2026-08-13
Double-quote all arguments in compiler/linker wrappers on Windows.

mmuetzel:
Apparently, Windows doesn't know how to deal with arguments in `args` vectors out of the box if they contain spaces.
That quirk is documented behavior:
https://learn.microsoft.com/en-us/cpp/c-runtime-library/spawn-wspawn-functions?view=msvc-170

Manually surround all arguments with double-quotes on Windows to deal with that quirk.

Do not add an additional set of double quotes when echoing the command in the wrappers on Windows.

Additionally, use the `exec_compiler` function that deals - among other things - with these platform specific quirks also in the `elmerld` wrapper.

### DIFF
--- a/fem/src/elmerf90-helper.in.h
+++ b/fem/src/elmerf90-helper.in.h
@@ -58,7 +58,17 @@ static void push(const char *s)
         fprintf(stderr, "elmerf90: too many arguments\n");
         exit(1);
     }
+
+#if defined (_WIN32)
+    // The spawn* functions on Windows require that their arguments are
+    // surrounded by double-quotes (to deal with spaces in arguments).
+    // See: https://learn.microsoft.com/en-us/cpp/c-runtime-library/spawn-wspawn-functions?view=msvc-170
+    args[nargs] = malloc(strlen(s) + 3);
+    sprintf(args[nargs], "\"%s\"", s);
+    nargs++;
+#else
     args[nargs++] = strdup(s);
+#endif
 }
 
 /* Split a whitespace-separated flag string and push each token. */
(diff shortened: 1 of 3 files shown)


king-aj3:
Tested #865 on the same machine and the same way as the original report — the
nompi zip from run 31708740725, everything driven from subprocesses with exit
codes read directly. **It fixes every case I had, including ones I did not
report.**

Same tree reached through junctions so only the path differs:

| case | before | with #865 |
|---|---|---|
| `ELMER_HOME` without a space (control) | exit 0, 80,070 B | exit 0, 80,070 B |
| `ELMER_HOME` **with** a space | exit 1, no dll | **exit 0, 80,070 B** |
| clean prefix, **source dir** with a space | exit 1, no dll | **exit 0, 80,070 B** |
| `…\Program Files\Elmer 26.2-c2a438a85` **and** case dir `My Cases\run one` | not previously tested | **exit 0, 80,070 B** |

That last row is @tmyllaeri's actual reported shape — a space in the parent and
in the leaf, plus two more in the working directory — which is a harder case
than anything in my first report. It builds.

The output `udf.dll` is 80,070 bytes in every passing row, byte-identical to
the no-space control, so this is not a path that merely stops erroring.

No regression on the control, and the run still reports the PATH fallback
(`'D:/a/_temp/msys64/ucrt64/bin/gfortran.exe' not usable … falling back to
'gfortran.exe' on PATH`) exactly as before, so #862's behaviour is intact.

For completeness on the point you raised: I have only measured Windows. The
`exec*`-family reading matches what I saw — the failure reproduced through the
**primary** `_spawnvp` with no fallback taken, so it was never specific to the
#862 fallback path.

Thanks for turning this round so quickly.

mmuetzel:
One of the macOS runners failed:
```
	607 - StressConstraintModes3 (Failed)                   cmodes eigen failing serial
```

Re-running the test passed without issues.
Maybe, an effect of removing `OMP_NUM_THREADS=1` from that workflow recently. But I don't think it can be related to the changes from this PR (which only affect Windows).

PR #867, opened 2026-08-18
Cast argv to the CRT's declared type in the Windows compiler/linker wrappers

### LINKED ISSUE #866: Windows MinGW/UCRT64 build failure with GCC >= 14: _spawnvp argv pointer type in elmerf90-helper.in.h
Building the Windows compiler/linker wrappers on MSYS2 UCRT64 fails in both `elmerf90.c` and `elmerld.c`. Reproduced against `devel` at 443882e04.

### Error

```
elmerf90-helper.h:146:39: error: passing argument 3 of '_spawnvp' from incompatible pointer type [-Wincompatible-pointer-types]
  146 |     int status = _spawnvp(P_WAIT, fc, args);
      |                                       ^~~~
      |                                       |
      |                                       char **
C:/msys64/ucrt64/include/process.h:89:88: note: expected 'const char * const*' but argument is of type 'char **'
   89 |   _CRTIMP intptr_t __cdecl _spawnvp(int _Mode,const char *_Filename,const char *const *_ArgList);
      |                                                                     ~~~~~~~~~~~~~~~~~~~^~~~~~~~
elmerf90-helper.h:161:45: error: passing argument 3 of '_spawnvp' from incompatible pointer type [-Wincompatible-pointer-types]
  161 |             status = _spawnvp(P_WAIT, base, args);
      |                                             ^~~~
      |                                             |
      |                                             char **
```

`-Wincompatible-pointer-types` is an error rather than a warning as of GCC 14, so on those toolchains this stops the build outright instead of just adding noise. Older GCC still builds.

### Affected

- MSYS2 / MinGW-w64 UCRT64, GCC >= 14 -- verified here on GCC 16.1.0; any MinGW toolchain with GCC >= 14 should hit the same error
- `fem/src/elmerf90-helper.in.h`, lines 146 and 161 at 443882e04
- Windows branch only; the POSIX `execvp()` branch is unaffected

### Cause

UCRT declares the third argument as `const char *const *`:

```c
intptr_t _spawnvp(int _Mode, const char *_Filename, const char *const *_ArgList);
```

The wrapper builds its argv as `static char *args[MAX_ARGS]`, i.e. `char **`, and passes it straight through. This is the same mismatch already handled for `_execvp` at the top of the same file, where the compatibility macro casts argv; the `_spawnvp` call sites just miss the cast.

### Patch

```diff
--- a/fem/src/elmerf90-helper.in.h
+++ b/fem/src/elmerf90-helper.in.h
@@ -143,7 +143,8 @@ static int exec_compiler(const char *fc, const char *who)
 
 #if defined (_WIN32)
     /* Spawn new process and wait for its exit code on Windows. */
-    int status = _spawnvp(P_WAIT, fc, args);
+    /* _spawnvp wants argv as 'const char *const *', like _execvp above. */
+    int status = _spawnvp(P_WAIT, fc, (const char * const *)args);
     if (status == -1) {
         int first_errno = errno;
 
@@ -158,7 +159,7 @@ static int exec_compiler(const char *fc, const char *who)
 
             free(args[0]);
             args[0] = strdup(base);
-            status = _spawnvp(P_WAIT, base, args);
+            status = _spawnvp(P_WAIT, base, (const char * const *)args);
             if (status == -1)
                 fprintf(stderr,
                         "%s: could not exec build-time compiler '%s' (%s), "
```

It is a cast only, so there is no runtime effect: the value passed is unchanged, only its type.

### Verification

MSYS2 UCRT64, GCC 16.1.0, against `devel`:

- without the casts, `elmerf90.c` and `elmerld.c` both fail with the two errors above
- with the casts, both compile clean and the build completes

Happy to open a PR with this if that is preferred.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

king-aj3:
Fixes #866.

`process.h` declares `_spawnvp`'s argv parameter as `const char *const *`, while the wrapper helper builds `args` as `char **`. The `-Wincompatible-pointer-types` warning is elevated to an error by default with GCC 14 or later, so building the wrappers fails on MSYS2 UCRT64 unless the diagnostic is downgraded — which is how CI currently avoids it (`-DCMAKE_C_FLAGS="-Wno-error=incompatible-pointer-types"`, added for Zoltan).

Two commits:

1. Apply the same cast the header's `execvp()` shim used, at both `_spawnvp` call sites. One change covers `elmerf90` and `elmerld`, since both include the helper. No runtime behaviour change — the CRT never writes through argv.
2. Drop the now-unused Windows `execvp` shim (the Windows branch calls `_spawnvp` directly; only the POSIX branch calls `execvp`), keeping `<process.h>` for `_spawnvp`.

Verified on MSYS2 UCRT64 (gcc 16.2.0, mingw-w64 headers 14.0.0.r262): the patched call shape compiles clean with `-Werror=incompatible-pointer-types` forced, at `-std=gnu11`, `gnu17` and `gnu23`.

### DIFF
--- a/fem/src/elmerf90-helper.in.h
+++ b/fem/src/elmerf90-helper.in.h
@@ -143,7 +143,12 @@ static int exec_compiler(const char *fc, const char *who)
 
 #if defined (_WIN32)
     /* Spawn new process and wait for its exit code on Windows. */
-    int status = _spawnvp(P_WAIT, fc, args);
+    /* _spawnvp wants argv as 'const char *const *'; our argv is built
+       as 'char **', and the mismatch warning is elevated to an error
+       by default with GCC 14 or later. Cast at the call site rather
+       than changing argv's type, since push()/args are shared with
+       other array/string logic. */
+    int status = _spawnvp(P_WAIT, fc, (const char * const *)args);
     if (status == -1) {
         int first_errno = errno;
 
@@ -158,7 +163,7 @@ static int exec_compiler(const char *fc, const char *who)
 
             free(args[0]);
             args[0] = strdup(base);
-            status = _spawnvp(P_WAIT, base, args);
+            status = _spawnvp(P_WAIT, base, (const char * const *)args);
             if (status == -1)
                 fprintf(stderr,
                         "%s: could not exec build-time compiler '%s' (%s), "
--- a/fem/src/elmerf90-helper.in.h
+++ b/fem/src/elmerf90-helper.in.h
@@ -4,12 +4,7 @@
  */
 
 #ifdef _WIN32
-#  include <process.h>   /* _execvp */
-   /* _execvp wants argv as 'const char *const *'; our argv is built as
-      'char **', which POSIX execvp() accepts directly but _execvp() does
-      not. Cast explicitly at the call site rather than changing argv's
-      type, since push()/args are shared with other array/string logic. */
-#  define execvp(prog, argv) _execvp((prog), (const char * const *)(argv))
+#  include <process.h>   /* _spawnvp */
 #else
 #  include <unistd.h>    /* execvp */
 #endif


fem/src/elmerf90-helper.in.h
```diff
@@ -143,7 +143,10 @@ static int exec_compiler(const char *fc, const char *who)
 
 #if defined (_WIN32)
     /* Spawn new process and wait for its exit code on Windows. */
-    int status = _spawnvp(P_WAIT, fc, args);
+    /* Cast as the execvp() shim above does: the CRT declares argv as
+       'const char *const *', and GCC 14/15 promote the mismatch to an
```
mmuetzel: Reading this, it sounds like the behavior changed again in GCC 16. That is not the case as far as I can tell.
Maybe, rephrase to make it clear that the warning is elevated to an error by default with GCC 14 or later.

mmuetzel: Same also goes for the commit message which sounds slightly misleading imho.

mmuetzel:
While at it, maybe remove the `#define execvp` on Windows (which is no longer used) in a separate commit. (I missed that in previous changes.)

fem/src/elmerf90-helper.in.h
mmuetzel: You could potentially use the existing phrasing for the comment about `_execvp` (before removing it in a followup commit) and just replace the function name with `_spawnvp`. That comment is pretty clear and to the point imho.

king-aj3:
All three addressed:

* the comment now reuses the shim's phrasing, with the compiler point stated as "elevated to an error by default with GCC 14 or later" — same rewording in the commit message, and the PR description is updated to match;
* the casts are unchanged at both call sites;
* the dead Windows `execvp` shim is removed in a separate commit, keeping `<process.h>` for `_spawnvp`.

mmuetzel:
Thanks for the follow-ups.

This looks good to me now. 👍

@raback: OK to merge?

raback:
Thanx for the contribution king and Markus for the review!

PR #872, opened 2026-08-20
AdvectionDiffusion: fix the Density scaling of the migration and Soret extra terms

### LINKED ISSUE #870: `AdvectionDiffusion`: the extra-term coefficients are scaled by `Density` in the wrong unit mode — migration and Soret are broken in *opposite* modes, so neither mode is correct for both
## Summary

`AdvectionDiffusion` carries two optional terms beyond convection-diffusion: Nernst–Planck **electromigration** (`Ion Charge`) and **thermodiffusion** (`Soret Diffusivity`). Each is silently scaled by an extra factor of `Density` — but *in opposite concentration-unit modes*:

| | relative mass (default) | `Concentration Units = "Absolute Mass"` |
|---|---|---|
| electromigration | **wrong by ×ρ** | correct |
| Soret / thermodiffusion | correct | **wrong by ×ρ** |

**There is therefore no setting in which both are correct.** Any thermo-electrochemical model — an electrochemical cell with Joule heating is the ordinary case — is wrong in one term whichever mode is chosen. Nothing errors and nothing warns; both produce converged, smooth, entirely plausible fields.

For a liquid electrolyte ρ ≈ 10³, so the affected term is wrong by three orders of magnitude.

## A1 — electromigration, wrong in relative-mass (default) units

`fem/src/modules/AdvectionDiffusion.F90`:

| line | code |
|---|---|
| 633 | `C2(i,j,:) = Density * Diffusivity(i,j,:)` (relative-mass branch) |
| 640 | `C2(i,j,:) = Diffusivity(i,j,:)` (absolute-mass branch) |
| 1340 | `Velo(i) = Velo(i) - C2(i,j) * NEConst * GradPhi(j)` (`DiffuseConvectiveCompose`) |
| 2014 | same line in `DiffuseConvectiveGenCompose` |
| 1439 | `A = A + C1 * Velo(i) * dBasisdx(q,i) * Basis(p)` |

`Velo` is a **velocity** and the convection matrix multiplies it by `C1`. Per the header comment at lines 112–114:

```fortran
! Relative mass units: C1=Density, C2=Density*Diff, C0 = 0, Ct=C1
! Absolute mass units: C1=1,       C2=Diff,         C0 = div v, Ct=1
```

so the assembled drift is `C1*C2*NEConst*grad(phi)`:

* absolute mass: `1 * D * NEConst` — correct
* relative mass: `rho * (rho D) * NEConst` — **one factor of `Density` too many**

The migration velocity should be built from the bare `Diffusivity`, not `C2`.

### Reproducer A1

Steady 1-D diffusion + migration, no flow, closed form:

```
C(y) = [exp(-beta y/Ly) - exp(-beta)] / [1 - exp(-beta)]
beta = z (e / kB T) (phi(Ly) - phi(0))
```

`Ly = 1 mm`, `D = 1e-9 m^2/s`, `z = +1`, `T = 298.15 K`, `dphi = 0.05 V`, `Density = 1200` ⇒ exact `beta = 1.9461`.

```
Constants
  Unit Charge = Real 1.602176634e-19
  Boltzmann Constant = Real 1.380649e-23
End
Equation 1
  Active Solvers(2) = 1 2
  Convection = Constant
End
Material 1
  Density = Real 1200.0
  Electric Conductivity = Real 10.0
  Conc Diffusivity = Real 1.0e-9
  Reference Temperature = Real 298.15
  Conc EPotential Field Variable = String "Potential"
  Conc Ion Charge = Real 1.0
  Convection Velocity 1 = Real 0.0
  Convection Velocity 2 = Real 0.0
End
! Solver 1 = StatCurrentSolve (Variable "Potential")
! Solver 2 = AdvectionDiffusion (Variable "Conc")
! BC y=0  : Potential = 0.0,  Conc = 1.0
! BC y=Ly : Potential = 0.05, Conc = 0.0
```

Result, `max |C - C_exact|` on a field whose range is `[0,1]`:

| concentration units | error | |
|---|---|---|
| relative (default) | **1.070e+00** | wrong |
| `Absolute Mass` | 7.598e-08 | correct |
| relative, with `Ion Charge = z/rho` | 1.930e-08 | correct (workaround) |

The factor is exactly `Density`, confirmed by fitting `beta` while sweeping the drive:

| `dphi` | `T` (K) | `beta` fitted | `beta` exact | ratio |
|---|---|---|---|---|
| 1.0e-3 V | 298.15 | 46.7151 | 0.0389 | **1200.23** |
| 1.0e-4 V | 298.15 | 4.6706 | 0.0039 | **1200.00** |
| 1.0e-3 V | 596.30 | 23.3542 | 0.0195 | **1200.06** |

Constant to **0.02 %** across a 10× voltage change *and* a 2× temperature change, and equal to the material `Density`. Both `phi` and `T` cancelling also confirms the `1/T` in `NEConst` is applied correctly.

## A2 — Soret / thermodiffusion, wrong in absolute-mass units

Same file:

| line | code |
|---|---|
| 660 | `C0 = Density` — **unconditional, both unit modes** |
| 1230 | `NodalCThermal = NodalC0(1:n)` |
| 1237 | `NodalC0 = 0.0d0   ! this is the way it works, maybe could do better some time` |
| 1491 | `SoretForce = CThermal * SorD * SUM( GradTemp(1:dim) * dBasisdx(p,1:dim) )` |
| 2192 | same in `DiffuseConvectiveGenCompose` |

`C0` is overloaded as the carrier for the extra-term coefficient, but it is loaded with `Density` regardless of unit mode. That is right for relative mass and wrong by `Density` for absolute mass, where the species equation has `C1 = Ct = 1`.

### Reproducer A2

Quasi-1-D strip, **no flow**, a linear temperature across the gap, concentration Dirichlet at `y=0` and **natural (zero flux)** at `y=Ly`. At steady state the total flux vanishes, so

```
-D dC/dy + D_T dT/dy = 0   ->   dC/dy = (D_T/D)(dT/dy) = constant
```

With `D = 1e-9`, `D_T = 5e-12`, `Ly = 1 mm`, `T: 300 -> 400 K`, `Density=1000`: `|dC/dy| = 500 /m`, so `C(Ly)` must be **0.5** with `C(0) = 1`.

```
Material 1
  Density = Real 1000.0
  Heat Conductivity = Real 0.6
  Heat Capacity = Real 4000.0
  Conc Diffusivity = Real 1.0e-9
  Conc Soret Diffusivity = Real 5.0e-12
  Conc Temperature Field Variable = String "Temperature"
  Reference Temperature = Real 300.0
  Convection Velocity 1 = Real 0.0
  Convection Velocity 2 = Real 0.0
End
! Solver 1 = HeatSolve, Solver 2 = AdvectionDiffusion
! BC y=0  : Temperature = 300, Conc = 1.0
! BC y=Ly : Temperature = 400          (no Conc BC -> natural)
```

Measured `C(Ly)`:

| concentration units | `Soret Diffusivity` absent | present | |
|---|---|---|---|
| relative (default) | 1.00000 | **0.50000** | correct |
| `Absolute Mass` | 1.00000 | **-499.00000** | **wrong** |

`ΔC` is `-0.5` in relative mass and `-500` in absolute mass: a ratio of exactly **1000 = `Density`**. Control: the temperature field genuinely varies 300 → 400 K in both runs, so "inert" is not a possible misreading.

## Related observation: dead code

`NodalCNP` is declared (1204, 1837) and assigned (`NodalCNP = NodalC0`, 1235, 1877) but **never read anywhere**. It looks like an unfinished attempt to give electromigration the same `C0`-carrier treatment Soret has — which may be why the drift ended up folded through `C2` instead, producing A1.

## No regression coverage for either term

Across `fem/tests`, `.sif` files containing:

| keyword | count |
|---|---|
| `Ion Charge` | **0** |
| `Soret Diffusivity` | **0** |
| `Diffusivity` (control) | 13 |

Both extra terms are entirely untested upstream, which is presumably how two factor-of-ρ errors survived.

## Suggested direction

Offered as a direction, not a verified patch — the maintainers will know the intended convention better than I do.

* **A1:** `NEConst` is computed once (503–513) and passed to both compose routines, so a single correction after `Density` is read (line 540) fixes both call sites: `IF ( .NOT. AbsoluteMass ) NEConst(1:n) = NEConst(1:n) / Density(1:n)`. Dividing by `C1` at the point of use is **not** equivalent: `C1` is set to `0.0` at line 626 when there is neither convection nor migration.
* **A2:** make the `C0` carrier unit-aware at line 660 — `Density` when `.NOT. AbsoluteMass`, `1.0` otherwise.
* **Tests:** both reproducers above are cheap (a 2×400 quad strip, direct solver, well under a second) and have closed-form references, so they make reasonable `fem/tests` cases with a `Reference Norm`. **Assert both directions** — with the keyword absent, `Ion Charge` sets `NEConst = 0` (506–508) and migration is silently *off*, so a test that only checks "the term fires" will not catch the scaling, and one that only checks the scaling will not catch a keyword that stopped being read.

I'm happy to open a PR with the fixes plus the two regression tests.

Foadsf:
Fixes #870.

`AdvectionDiffusion` carries two optional extra terms beyond convection-diffusion: Nernst-Planck electromigration (`Ion Charge`) and thermodiffusion (`Soret Diffusivity`). Each is silently scaled by a surplus factor of `Density`, but **in opposite concentration-unit modes**, so no setting is correct for both. Nothing errors and nothing warns.

|  | relative mass (default) | `Concentration Units = "Absolute Mass"` |
|---|---|---|
| electromigration | **wrong by x rho** | correct |
| Soret | correct | **wrong by x rho** |

## The changes

**Electromigration.** The drift velocity is built from `C2` in the compose routines (`Velo = Velo - C2*NEConst*grad(phi)`) and the convection matrix then multiplies it by `C1`. In relative mass units `C1 = Density` and `C2 = Density*Diffusivity`, so the assembled term carries `Density` twice. The surplus factor is removed from `NEConst` where the unit mode is already known.

Dividing by `C1` at the point of use is **not** equivalent: `C1` is set to `0` when there is neither convection nor migration. The `WHERE` guard exists only to preserve today's behaviour when `Density` is absent or zero, where the relative-mass formulation is degenerate (`C1 = CT = 0`) and a bare division would produce a NaN. Happy to drop it if you would rather it were a plain division.

**Soret.** `C0` is used only as the carrier for the thermal diffusion coefficient -- the compose routines copy `NodalC0` into `NodalCThermal` and then zero `NodalC0` unconditionally, so `C0` at the integration point is always `0` and never acts as a reaction term -- but it was loaded with `Density` regardless of unit mode. It is now unit-aware. The `ScaledToSolubility` division is unchanged and still applied afterwards.

## Verification

Both terms have closed-form references. Measured on a 1D strip against exact solutions:

| case | before | after |
|---|---|---|
| migration, relative mass | drift ratio **7835.85**, max err 1.4840 | ratio **1.00**, max err 4.3e-07 |
| migration, absolute mass | ratio 1.00 | ratio 1.00 (unchanged) |
| migration, drive sweep (0.05 / 1e-3 / 1e-4 V) | 1201.4 / 1200.0 ratio | **1.00** throughout, err ~5e-09 |
| Soret, absolute mass | C(Ly) = **-499.00000** | C(Ly) = **0.50000** |
| Soret, relative mass | 0.50000 | 0.50000 (unchanged) |

The `Density` factor was confirmed by fitting the drift exponent while sweeping the drive: constant to 0.02 % across a 10x voltage change *and* a 2x temperature change, and equal to the material `Density`.

The two fixes are independent -- reverting either one alone breaks only its own term:

| build | migration (relative) | Soret (absolute) |
|---|---|---|
| both fixes | correct | correct |
| migration fix reverted | **wrong** | correct |
| Soret fix reverted | correct | **wrong** |

**Full suite: 908 / 911 pass.** The three failures (`EM_port_eigen_3D_prisms`, `JouleHeatToOpenFOAM`, `TemperatureFromOpenFOAM`) are pre-existing in this configuration -- they fail identically with the patch reverted and the module rebuilt, and `AdvectionDiffusion` does not appear in any of their logs. Built with gfortran 15.2.0, `WITH_MPI=FALSE`, on devel `48bde6c`.

## Tests

`Ion Charge` and `Soret Diffusivity` appeared in **no** `.sif` file under `fem/tests`, which is presumably how both errors survived. Four cases are added, one per (term, unit mode), each about half a second:

- `AdvDiffMigration`, `AdvDiffMigrationAbsMass`
- `AdvDiffSoret`, `AdvDiffSoretAbsMass`

Both unit modes are covered deliberately. The terms were wrong in *opposite* modes, so testing only the broken halves would leave the correct halves free to regress. Each reference norm is sensitive in both directions: it differs from the profile obtained when the keyword stops being read (linear for migration, flat for Soret) as well as from the density-scaled profile. Verified by reverting each fix and confirming exactly the corresponding test fails.

Reference norms were checked against the closed forms rather than merely recorded from the run: they agree to 2e-06 on the exponential profile and to ~1e-09 on the linear ones.

## Two notes for maintainers

1. **This changes results for anyone using the documented workaround.** Setting `Ion Charge = z/rho` in relative mass units compensated for the bug; with the fix it under-scales the drift by `1/rho` and must be removed. Measured, not predicted. Probably worth a release-note line.

2. `NodalCNP` / `CNP` is dead code (declared, assigned from `NodalC0`, never read) in both compose routines. It looks like an unfinished attempt to give electromigration the same carrier treatment Soret has, which may be why the drift ended up folded through `C2`. Left untouched here rather than folding an unrelated deletion into a bug fix -- happy to remove it in a follow-up if you want.

Separately, the header comment at the top of the solver still documents `C0 = div v` for absolute mass units, which the code has never done; `C0` is zeroed before use. Not changed here.


raback:
Thanx Foad! There seems to have been a bug for many years for Soret diffusivity. Good to have some new tests!

Issue #875, opened 2026-08-21
Symbol ModelLumping is used for module name and local variable, leading to compilation errors


lucas-ige:
Since [this commit](https://github.com/ElmerCSC/elmerfem/commit/0eb84ac73962bc773f42ce1c76350d452ca84731), the symbol `ModelLumping` is used both as the name of a module and a local variable in the routine where the module is used, which is illegal in FORTRAN AFAIK. At any rate, Elmer does not compile for me.

This issue occurs in:

 - `fem/src/modules/ElasticSolve.F90 `
 - `fem/src/modules/StressSolve.F90 `

I suggest to rename the logical variable `ModelLumping` into `UseModelLumping` in both cases. It fixes the compiling errors for me.

I will prepare the corresponding PR.

lucas-ige:
Fixed by PR #876.

PR #876, opened 2026-08-21
Fix symbol conflict (ModelLumping) in ElasticSolve.F90 and StressSolve.F90

lucas-ige:
This symbol is used both as a module name and a local variable.

This commit renames the local variable to avoid the conflict.

This commit fixes issue #875.

### DIFF
--- a/fem/src/modules/ElasticSolve.F90
+++ b/fem/src/modules/ElasticSolve.F90
@@ -455,7 +455,7 @@ SUBROUTINE ElasticSolver( Model, Solver, dt, TransientSimulation )
   ! the boundary. State of the run, deliberately NOT in any SAVE list -- it has to
   ! live across the six cases of THIS call and no longer, which is exactly the
   ! mistake PrevSOL used to make.
-  LOGICAL :: ModelLumping
+  LOGICAL :: UseModelLumping
   TYPE(ModelLumping_t) :: Lump
 
   ! Gates for the refusal of StressSolve-only keywords: true when one is set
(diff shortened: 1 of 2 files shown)


juharu:
After the checks finish, i'll pull. Just out of curiosity, which compiler are you using ? no indication from any compiler i've used recently (or github CI).

lucas-ige:
Thank you @juharu.

I was indeed surprised that none of the CI complained. I am using intel v20.0.0 but I can reproduce the problem with gfortran on a MWE:

```
$> cat mymod.f90 test.f90 
module mymod

contains

  subroutine mysub()
    implicit none
    print*,"Hello world"
  end subroutine mysub
  
end module mymod
program test

  use mymod, only: mysub
  
  implicit none
  
  logical :: mymod
  
  call mysub()
  
end program test

$> gfortran mymod.f90 test.f90
test.f90:7:18:

    7 |   logical :: mymod
      |                  1
Error: Symbol ‘test’ at (1) cannot have a type

$> ifort mymod.f90 test.f90
test.f90(7): error #6450: This name has already been used as an external module name.   [MYMOD]
  logical :: mymod
-------------^
compilation aborted for test.f90 (code 1)
```

gfortran is even confused about the symbol that creates the problem -- it tells me about `test` instead of `mymod` (tested gfortran v8.5.0 and gfortran v14.1.0). In this case intel gives me more useful information.

juharu:
Thanks. Might be that my and CI compilers are too old? Although, I think that one of mac runners even uses a Intel compiler. But doesn't much matter, your changes are fine anyway. I'll do the honours (after the checks all green).

juharu:
Thanks Lucas!

PR #880, opened 2026-08-24
Enable using distributed right-hand side of MUMPS 5.3 on all platforms

mmuetzel:
MUMPS has been rebuilt in MSYS2 with `-DAVOID_MPI_IN_PLACE`. That means using its distributed right-hand side should work correctly now. That MUMPS features has not been used in Elmer by default on Windows because it did not work with MUMPS that was built without that flag.

Enable the distributed right-hand side of MUMPS for all platforms by default.

Adapt the comment to describe the current situation and give a motivation for keeping the configuration flag that can be used to disable using the distributed right-hand side of MUMPS 5.3 or later.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -633,36 +633,39 @@ IF(Mumps_FOUND)
     ENDIF()
   ENDFOREACH()
 
-  # Off on Windows, because MUMPS's own reductions cannot work there. MPI_IN_PLACE
-  # is unusable with MS-MPI and gfortran: mpif.h keeps the sentinel in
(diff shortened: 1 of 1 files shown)


mmuetzel:
The CI still fails for MSYS2. The revert of the recent changes for msmpi is still in their staging repository only.

mmuetzel:
The reverted msmpi package is out of the staging repository now.

Marking as ready for review again.

juharu:
Tried  the new mumps on my laptop, worked out just fine. Thanks!

PR #884, opened 2026-08-26
Add matrix bulk assembly when using Hypre

dancaseb:
Bulk assembly of a matrix is now default when using Hypre. Row by row assembly (used previously) can be enabled by the sif parameter Hypre Assemble Row By Row = Logical True

The reason to rewrite this was that when running the code on GPUs, the runtime was too high due to so many CPU-GPU memory transfers.

Also added HYPRE_Init() calls and commented out the AMS coordinate vectors as it's not used and would trigger internal HYPRE errors.

Note this so far works for 1 GPU, not multi GPU. Needs to be investigated more.


juharu:
Ready to be merged, i guess?

raback:
Nice work!

PR #886, opened 2026-08-27
ElmerGUI: Adjust starting processes with Qt6

mmuetzel:
Windows seems to be particular picky when it comes to quotation rules for processes started with `QProcess`.
In particular, it refuses to run an executable if it isn't double-quoted. See: https://www.elmerfem.org/forum/viewtopic.php?p=33284#p33284

I'm suspecting that other platforms don't care as much about that. Additional double-quotes shouldn't hurt there - I guess. But it might make sense if someone could double-check that before merging this.

While at it, I also changed the default value for mpiexec that is used in the GUI on Windows to include the full path for the default installation location of it. That should avoid the need to manually add that path to the environment variable `PATH` and should make it work "out of the box" for the vast majority of GUI users on Windows. (That's by my personal estimate. I don't have any statistics to substantiate that claim. But from my experience, most Windows users just click "Next", "Next", "Next", ... in installers without caring too much what could be changed.)
Please, let me know if you prefer to keep the current setting. I don't have strong feelings either way about that part.

### DIFF
--- a/ElmerGUI/Application/src/mainwindow.cpp
+++ b/ElmerGUI/Application/src/mainwindow.cpp
@@ -48,6 +48,7 @@
 #include <QStringList>
 #include <QSystemTrayIcon>
 #include <QTimeLine>
+#include <QtGlobal>  /* for QT_VERSION_CHECK with Qt4 or Qt5 */
 #include <QtGui>
 
 #include <fstream>
@@ -5106,7 +5107,11 @@ void MainWindow::showVtkPostSlot() {
 
     logMessage("Executing: " + unifyingCommand);
 
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+    meshUnifier->startCommand(unifyingCommand);
+#else
     meshUnifier->start(unifyingCommand);
+#endif
 
     if (!meshUnifier->waitForStarted()) {
       solverLogWindow->getTextEdit()->append(
@@ -6638,7 +6643,11 @@ void MainWindow::runsolverSlot() {
 
       logMessage("Executing: " + partitioningCommand);
 
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+      meshSplitter->startCommand(partitioningCommand);
+#else
       meshSplitter->start(partitioningCommand);
+#endif
 
       if (!meshSplitter->waitForStarted()) {
         logMessage("Unable to start ElmerGrid for mesh partitioning - aborted");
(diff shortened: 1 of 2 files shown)


mmuetzel:
The forum user kevinarden pointed out that there is a file with the following text on the funet server:

> The Windows zip and exe packages on the funet mirror do not currently include the MinGW runtime DLLs that ElmerSolver.exe links against, so the installation procedure following the documented steps fails silently. At some stage the updated cmake recipes have been broken and the library modules are no longer bundled.
> 
> The nightly build was based on very old setup and was missing many important libraries (like MUMPS). Hence we will not try to repair the old nightly build. Here we provide an old still working version from July 1st 2026 (ignore the date).
> 
> A fresh version may be obtained from GutHub actions. Go to the following page:
> https://github.com/ElmerCSC/elmerfem/actions/workflows/build-windows-mingw.yaml
> 
> There choose a zip or exe package (with or without MPI) from the bottom of the page. These come with all the necessary files. Note that you need to be logged into GitHub to see the files available for download.

I don't know who curates that server (or how to ping on the forum). Maybe, @raback or @juharu?
I hope it is ok to bring that up here.
It might make sense to filter that link in the message for builds on the **devel** branch. Otherwise, it might become difficult to correlate error reports.
I.e.: https://github.com/ElmerCSC/elmerfem/actions/workflows/build-windows-mingw.yaml?query=branch%3Adevel

Also there is a typo in the line above the link: "GutHub" instead of "GitHub".

mmuetzel:
Wait a moment.

I took a look at the Qt6 documentation for QProcess::start:
https://doc.qt.io/qt-6/qprocess.html#start
Rather then using that, ElmerGUI should probably be using QProcess::startCommand on Qt6:
https://doc.qt.io/qt-6/qprocess.html#startCommand

That could also explain why users reported that it was working with the version that they got from the funet server. But not with the version they got from the workflow build artifacts.
The latter is build with Qt6. I assume the former might have been built with Qt5.

I'll revise the change to take that into account.
Converting to draft for the time being.

mmuetzel:
While going through `mainwindows.cpp` looking for how it uses `QProcess::start`, I noticed that it was still attempting to use the Windows .bat file for compilation. That has been replaced with a compiler wrapper executable a couple weeks (or is it already months?) ago.

That should also be addressed now.

raback:
Thanx Markus! I updated the info on nic.funet.fi. I can confirm that the old Windows installer was using Qt5. Good to have also this updated.

PR #888, opened 2026-08-28
Report real CPU time and memory on Windows

### LINKED ISSUE #887: bug: on Windows, CPUTime() and RealTime() are the same function, so every (CPU,REAL) pair is identical and cannot detect threading
## Summary

On Windows builds, `cputime()` and `realtime()` in `fem/src/CPUTime.c` are the **same function body**, so every `(CPU,REAL)` pair Elmer prints contains the same number twice. The practical consequence is that a user cannot tell from Elmer's own output whether an OpenMP build is threading at all — a genuinely 16-threaded run and a serial run both report `CPU == REAL`. `cpumemory()` on the same branch returns a hard-coded `0.0`.

I hit this while measuring whether enabling OpenMP was worth it, and initially concluded from Elmer's own timing that threads were **not** engaging. They were: the same runs are 3.15× faster and the OS reports 2.21 cores busy.

## Symptom

`fem/tests`-style case, 112k nodes / 547k tets, `StatCurrentSolveVec`, CG + ILU1, identical solution norm `0.65630454` throughout.

**Windows** (MSYS2 UCRT64, `WITH_OpenMP=TRUE`) — Elmer's own report:

| `OMP_NUM_THREADS` | solver time (CPU,REAL) | ratio |
|---|---|---|
| 1 | 10.04, 10.04 | 1.00 |
| 4 | 2.71, 2.71 | 1.00 |
| 16 | 2.23, 2.23 | 1.00 |

The wall time falls by 4.5×, so the threading is real — but CPU tracks it exactly, which CPU time cannot do.

Measured independently on the same binary via `Process.TotalProcessorTime`:

```
OMP_NUM_THREADS=1    CPU 11.56 s / wall 11.61 s  = 1.00 cores busy
OMP_NUM_THREADS=16   CPU 12.59 s / wall  5.70 s  = 2.21 cores busy
```

**Linux control** (Ubuntu 24.04, Elmer 26.2 from the PPA, same case, same mesh, same norm):

| `OMP_NUM_THREADS` | solver time (CPU,REAL) | ratio |
|---|---|---|
| 1 | 6.40, 6.39 | 1.00 |
| 4 | 6.40, 2.45 | **2.61** |

The POSIX path behaves correctly in both directions — it reports ~1.00 when the run genuinely is serial, and 2.61 when it is not. Only the Windows path is stuck at 1.00.

## Root cause

`fem/src/CPUTime.c`, lines 43–61:

```c
#if defined(MINGW32) || defined(WIN32)

double STDCALLBULL FC_FUNC(realtime,REALTIME) ( )
{
  return clock() / (double)CLOCKS_PER_SEC;
}

double STDCALLBULL FC_FUNC(cputime,CPUTIME) ( )
{
  return clock() / (double)CLOCKS_PER_SEC;
}

double STDCALLBULL  FC_FUNC(cpumemory,CPUMEMORY) ( )
{
  return 0.0;
}
```

against the POSIX branch below it, which correctly uses `getrusage(RUSAGE_SELF).ru_utime` for CPU and `gettimeofday` for wall.

Two identical bodies is the whole bug. On Windows the C runtime's `clock()` measures elapsed wall time since process start rather than consumed CPU time, so this branch returns wall time from both — which is what the measurements above show regardless of the mechanism.

## Reproduction

1. Configure with `-DWITH_OpenMP=TRUE` on MSYS2 UCRT64.
2. Any case whose solver has a threaded `*Vec` variant, with `Solver Timing = Logical True`.
3. Run at `OMP_NUM_THREADS=1` and again at 8 or 16.
4. Wall time drops; the reported CPU time drops with it, and the two are always equal.

## Suggested fix

For the Windows branch:

- `cputime()` → `GetProcessTimes()`, summing `lpUserTime` (optionally `lpKernelTime`) into seconds. That is the direct analogue of `getrusage(...).ru_utime` and sums across threads.
- `realtime()` → keep `clock()`, or `QueryPerformanceCounter` for better resolution.
- `cpumemory()` → `GetProcessMemoryInfo(...).WorkingSetSize`, the analogue of `ru_maxrss`.

All three are in `kernel32`/`psapi` and need no new dependency. Happy to send a PR if you would like one — I did not want to guess whether you would prefer `psapi` linkage for the memory part or would rather leave `cpumemory()` alone.

## Affected versions

- `devel` at `a38661b4b`, MSYS2 UCRT64, gfortran 15.2.0, Windows 11
- Elmer 26.2 (Ubuntu PPA) is **not** affected — the POSIX branch is correct

The Windows branch dates from the original 1997 file, so this is longstanding rather than a regression.

## Prevention

Two functions with identical bodies inside a platform `#if` is the tell — the POSIX branch differentiates them, the Windows one does not. Worth noting that this is a *diagnostic* that fails silently: it does not error, it reports a plausible number that quietly answers a different question, so it misleads precisely when someone is trying to find out whether their parallel build works.

Foadsf:
Fixes #887.

On Windows, `cputime()` and `realtime()` in `fem/src/CPUTime.c` had identical bodies — both `clock()/CLOCKS_PER_SEC` — and `cpumemory()` returned a hard-coded `0.0`. Because `clock()` measures elapsed wall time on Windows rather than consumed CPU time, every `(CPU,REAL)` pair Elmer prints contained the same number twice, so a threaded run was indistinguishable from a serial one in the solver's own output. The POSIX branch below already distinguishes them via `getrusage(RUSAGE_SELF)`.

## The change

| | before | after |
|---|---|---|
| `cputime()` | `clock()` — wall time | `GetProcessTimes()` user time, the analogue of `ru_utime`, summed across threads |
| `cpumemory()` | `0.0` | `GetProcessMemoryInfo()` `PeakWorkingSetSize`, the analogue of `ru_maxrss`, in kilobytes |
| `realtime()` | `clock()` | unchanged — already wall time here |

`cpumemory()` is scaled to kilobytes deliberately, to match both `ru_maxrss` and the `"maximum memory usage (kb)"` label `SaveScalars` prints.

**No build change is needed.** On MinGW-w64 the `psapi.h` declaration resolves to `K32GetProcessMemoryInfo` in `kernel32`, which is already linked — I checked this by building rather than assuming, since the alternative would have meant touching the Windows link line.

## Verification

Through `SaveScalars` (`Operator = cpu time / wall time / cpu memory`), same input, same mesh, 112k nodes / 547k tets:

```
before   cpu 10.727 s   real 10.727 s   memory      0 kb
after    cpu 11.375 s   real 11.592 s   memory 373404 kb
```

And the solver banner now tracks threading, where before it read 1.00 at every thread count while wall time fell 4.5×:

| `OMP_NUM_THREADS` | CPU/REAL after the fix |
|---|---|
| 1 | 0.99 |
| 4 | 2.15 |
| 16 | 3.45 |

That matches what Linux already reported for the same case — 1.00 serial, 2.61 at 4 threads — including the important half, that it reads ~1.0 when the run genuinely *is* serial rather than simply always exceeding 1.

`ctest` on MSYS2 UCRT64 with `WITH_OpenMP=TRUE`: **909/912**. The three failures — `EM_port_eigen_3D_prisms`, `JouleHeatToOpenFOAM`, `TemperatureFromOpenFOAM` — are identical to the pre-change baseline on the same machine and are unrelated to this change, which touches no numerics.

## Note

`GetProcessTimes` returns user time only, matching `ru_utime`. Kernel time is available as a separate `FILETIME` if you would rather the two platforms both reported user+system; I kept the narrower definition so the Windows and POSIX numbers mean the same thing. Happy to change it either way.

### DIFF
--- a/fem/src/CPUTime.c
+++ b/fem/src/CPUTime.c
@@ -43,7 +43,9 @@
 #if defined(MINGW32) || defined(WIN32) 
 
 #include <sys/types.h>
-#include <time.h> 
+#include <time.h>
+#include <windows.h>
+#include <psapi.h>
 
 double STDCALLBULL FC_FUNC(realtime,REALTIME) ( )
 {
@@ -52,12 +54,39 @@ double STDCALLBULL FC_FUNC(realtime,REALTIME) ( )
 
 double STDCALLBULL FC_FUNC(cputime,CPUTIME) ( )
 {
-  return clock() / (double)CLOCKS_PER_SEC;
+  /* User CPU time consumed by this process, summed over all its threads --
+     the analogue of getrusage(RUSAGE_SELF).ru_utime used by the POSIX branch
+     below.
+
+     clock() must not be used here.  On Windows it measures elapsed wall time
+     rather than consumed CPU time, so it returns the same value as realtime()
+     above, and every (CPU,REAL) pair Elmer prints becomes two copies of the
+     same number.  That makes a threaded run indistinguishable from a serial
+     one in the solver's own output. */
+  FILETIME created, exited, kernel, user;
+  ULARGE_INTEGER t;
+
+  if ( !GetProcessTimes( GetCurrentProcess(), &created, &exited, &kernel, &user ) )
+    return 0.0;
+
+  t.LowPart  = user.dwLowDateTime;
+  t.HighPart = user.dwHighDateTime;
+
+  /* FILETIME counts 100-nanosecond intervals. */
+  return (double)t.QuadPart * 1.0e-7;
 }
 
 double STDCALLBULL  FC_FUNC(cpumemory,CPUMEMORY) ( )
 {
-  return 0.0;
+  /* Peak working set, the analogue of getrusage(RUSAGE_SELF).ru_maxrss.
+     ru_maxrss is reported in kilobytes, so scale to match: the callers of
+     this function compare the two platforms' numbers. */
+  PROCESS_MEMORY_COUNTERS pmc;
+
+  if ( !GetProcessMemoryInfo( GetCurrentProcess(), &pmc, sizeof(pmc) ) )
+    return 0.0;
+
+  return (double)pmc.PeakWorkingSetSize / 1024.0;
 }
 
 #else


juharu:
Seems good to me, thanks!

juharu:
The CI  on devel failing windows builds atm, not related to this, but my other doings. I'll correct that after finishing the fix.

PR #890, opened 2026-08-30
ElmerGUI: link the ElmerGrid core instead of copying its sources

### LINKED ISSUE #889: ElmerGUI duplicates the ElmerGrid sources; egdef.h has drifted between the two copies
## Summary

`ElmerGUI/Application/plugins/` holds verbatim copies of four ElmerGrid source files and five of its headers. The copies are refreshed by hand with a checked-in script, `ElmerGUI/Application/plugins/grid2gui.sh`. That script does not copy `egdef.h`, and `egdef.h` has consequently drifted: the two locations currently define a different value of `FM_PI`.

Everything below is against `devel` at `4f2d7e4b9`.

## The duplication

`grid2gui.sh` is the whole synchronisation mechanism:

```bash
cp -f ../../../elmergrid/src/egconvert.c egconvert.cpp
cp -f ../../../elmergrid/src/egmesh.c egmesh.cpp
cp -f ../../../elmergrid/src/egnative.c egnative.cpp
cp -f ../../../elmergrid/src/egutils.c egutils.cpp
cp -f ../../../elmergrid/src/egconvert.h .
cp -f ../../../elmergrid/src/egmesh.h .
cp -f ../../../elmergrid/src/egnative.h .
cp -f ../../../elmergrid/src/egutils.h .
cp -f ../../../elmergrid/src/egtypes.h .
```

The four `.c` files are 21,454 lines in total and are currently byte-identical to their `.cpp` counterparts, which `git ls-tree` shows directly, as the same blob at two paths:

```
$ git ls-tree devel elmergrid/src/egutils.c ElmerGUI/Application/plugins/egutils.cpp
100644 blob a370879cf5ce18a0baeda71ccd8bdc4662c23bcd	ElmerGUI/Application/plugins/egutils.cpp
100644 blob a370879cf5ce18a0baeda71ccd8bdc4662c23bcd	elmergrid/src/egutils.c
```

## The drift

`egdef.h` is copied into the plugin directory but is **not** in `grid2gui.sh`, so running the script does not refresh it. The two copies have diverged:

```
elmergrid/src/egdef.h:34                   #define FM_PI 3.141592653
ElmerGUI/Application/plugins/egdef.h:16    #define FM_PI 3.1415926
```

`FM_PI` is used 25 times in `egmesh.c` and once in `egnative.c`, so ElmerGrid and the mesher built into ElmerGUI currently run the same code with a different value of pi. The difference is small (about 1.7e-8 relative) and I have not tried to show a visible effect on a mesh, but the two tools are meant to be the same code and here they are not.

The GUI copy of `egdef.h` is also older in a second way: it predates the `EG_PLUGIN` block that `elmergrid/src/egdef.h` now uses to select the embedded configuration.

```c
/* elmergrid/src/egdef.h */
#ifdef EG_PLUGIN
#define USE_MATC 0
#else
...
```

`ElmerGUI/Application/CMakeLists.txt:105` already does `ADD_DEFINITIONS(-DEG_PLUGIN)`, so the definition is set, but the header it reaches has no `EG_PLUGIN` handling and never defines `USE_MATC` or `USE_METIS` at all. `egnative.c` tests those 18 times as `#if USE_MATC` / `#if USE_METIS`. Undefined identifiers evaluate to `0` in `#if`, which happens to be the intended embedded configuration, so **the current build is not broken** — but it arrives there by accident rather than through the switch that was written for exactly this purpose, and it would warn under `-Wundef`.

## This has caused a real divergence before

- `85e16d005` (2026-02-26, "fixes #782") changed `elmergrid/src/egmesh.c` only, one file.
- `145defd27` (2026-07-05) is titled "ElmerGrid, fix compiler warnings from -Wmaybe-uninitialized -Wuninitialized. **Synchronize with ElmerGUI**."

So the fix in February reached ElmerGrid and did not reach the copy compiled into ElmerGUI until the resynchronisation in July. Commit messages such as "Make ElmerGUI code consistent with changes" and "Add tentative support also for ElmerGUI" show this is recurring maintenance rather than a one-off.

## Suggested fix

Build the shared routines once and link them, rather than copying them:

1. Add a static library target in `elmergrid/src` built from `egconvert.c`, `egmesh.c`, `egnative.c` and `egutils.c` with `EG_PLUGIN` defined.
2. Add an aggregate header wrapping the existing headers in `extern "C"`, since the core is C and ElmerGUI is C++. The individual headers mostly have no include guards, so one wrapper is less invasive than editing six files.
3. Have `ElmerGUI` link that target and delete the copies together with `grid2gui.sh`.

`ADD_SUBDIRECTORY(elmergrid)` at `CMakeLists.txt:828` is unconditional and precedes `ADD_SUBDIRECTORY(ElmerGUI)` at line 843, so the target always exists when ElmerGUI is configured. The `ElmerGrid` executable does not need to change.

I have this working on a branch and will open a pull request against `devel` referencing this issue. Happy to take a different approach if you would prefer the core library shaped another way, or if the duplication is deliberate for a reason I have missed.

## Environment

Reported against `devel` at `4f2d7e4b9`. The finding is in the repository contents and the CMake wiring, so it is not specific to an OS, compiler or MPI configuration.

Foadsf:
Closes #889.

`ElmerGUI/Application/plugins/` carried verbatim copies of `egconvert.c`, `egmesh.c`, `egnative.c` and `egutils.c` (renamed to `.cpp`) plus five shared headers, refreshed by hand with `plugins/grid2gui.sh`. Because that script does not copy `egdef.h`, that header had drifted: the ElmerGUI copy defines `FM_PI` as `3.1415926` where `elmergrid/src` defines `3.141592653`, and it predates the `EG_PLUGIN` block that selects the embedded configuration. The evidence is written up in #889.

This builds the shared routines once and links them instead.

## What changed

- `elmergrid/src/CMakeLists.txt` — new `elmergrid_core` static library from `egconvert.c`, `egmesh.c`, `egnative.c`, `egutils.c`, with `EG_PLUGIN` as a `PUBLIC` compile definition so the setting cannot drift between the library and ElmerGUI's own sources.
- `elmergrid/src/elmergrid_core.h` — new aggregate header wrapping the existing headers in `extern "C"`. The core is C, ElmerGUI is C++. Most of the individual headers carry no include guards, so one wrapper seemed less invasive than editing six of them; happy to change this if you would rather guard the headers themselves.
- `ElmerGUI/Application/CMakeLists.txt` — drops the nine duplicated files from `HEADERS`/`SOURCES` and links `elmergrid_core`.
- `ElmerGUI/Application/plugins/egmain.cpp` — includes `elmergrid_core.h` in place of the six individual headers. No other change.
- Deletes the nine copies and `grid2gui.sh`.

Net `+80 / −25275`.

`ADD_SUBDIRECTORY(elmergrid)` at `CMakeLists.txt:828` is unconditional and precedes `ADD_SUBDIRECTORY(ElmerGUI)` at line 843, so the target always exists when ElmerGUI is configured. **The `ElmerGrid` executable target is untouched** — no source, flag or link change.

## Behaviour

Intended to be none. ElmerGUI compiled these sources with `-DEG_PLUGIN` before and compiles the same sources with `-DEG_PLUGIN` now. The one deliberate difference is that ElmerGUI now sees the current `egdef.h`, so it picks up `FM_PI 3.141592653` and reaches `USE_MATC 0` / `USE_METIS 0` through the `EG_PLUGIN` branch rather than through those identifiers being undefined in `#if`. Both resolve to the same configuration; the pi value does change, in the direction of matching `ElmerGrid`.

The sources are now compiled as C for ElmerGUI rather than as C++. They are C, and `ElmerGrid` has always compiled them as C.

## What I verified

Local, gcc/g++ 13 on Ubuntu (WSL2):

- All eight `elmergrid/src` sources compile as C with `DISABLE_MATC` (the `ElmerGrid` configuration).
- The four core sources compile as C with `EG_PLUGIN` and archive into `libelmergrid_core.a`.
- `plugins/egmain.cpp`, `plugins/elmergrid_api.cpp` and `src/meshtype.cpp` compile as C++ against `elmergrid_core.h`, with no warnings from `egmain.cpp`.
- Those objects link against the archive into a running binary that constructs `ElmergridAPI` and calls into the core. As a control, rebuilding the archive from C++-compiled objects makes the same link fail with 50 undefined references, so the successful link is actually testing the `extern "C"` boundary.

## What I could not verify

- **I have not run `ctest`, and I have not built ElmerGUI itself.** The checks above were run in a Linux (WSL2) environment with no Qt, so I compiled and linked the changed translation units directly rather than configuring the GUI. The changed translation units and the C/C++ link boundary are exercised; the Qt-side build is not.

  *Correction to an earlier version of this description:* I first wrote that this machine has no Qt, VTK or QWT. That was true only of the Linux environment I ran the checks in — Qt and VTK are in fact installed on the Windows side of the same machine. So the accurate statement is that I did not attempt the full Elmer + ElmerGUI build, not that I could not. I am happy to run it and report the result if that would help the review.
- Not built on Windows or macOS. The change is CMake and include wiring, so I would not expect platform-specific behaviour, but I have not shown it.

Please push back if the library shape, its name, or the aggregate-header approach is not what you would want here.


Foadsf:
The failing `macos-15-intel (using openblas)` check here is not caused by this pull request. It fails in the `install dependencies` step, before anything is configured or compiled, with `vtk: no bottle available!` — Homebrew publishes no Intel macOS bottle for vtk (as of 9.7.0 the tags are arm64_tahoe, arm64_sequoia, arm64_sonoma, arm64_linux, x86_64_linux).

The same job and step fail on unrelated branches, including `fix/permafrost-hydraulic-reference-state` (run 33298600360) several hours before this branch existed, and `interfrost-devel` the day before. The last green runs of that workflow were on 2026-08-28.

Filed as #893 with a fix in #894. The `macos-14` legs pass here.

Foadsf:
I have temporarily cherry-picked the two commits from #894 onto this branch so that CI can actually run here. Without them the `macos-15-intel` leg dies in `install dependencies` on `vtk: no bottle available!` before anything in this pull request is compiled, as described above.

Those two `ci:` commits are **not part of this change** and are identical to the ones in #894 (same patches, same resulting workflow file). Once #894 is merged I will rebase this branch onto `devel`, which drops them automatically, leaving only the commit this pull request is actually about.

If you would rather I keep this branch free of them and simply leave the macOS check red until #894 lands, say so and I will drop them again.

richb2k:
Glad to see progress toward having a single copy of ElmerGrid.

IMHO here are two suggestions, please don't take the comments as required.
1.  Add the Extern C guards to each header file instead of adding a wrapper.  This reduces the number of source file by one, and makes things simpler.
2. Take a look at re-enabling including MATC in ElmerGrid in the cmake efforts, (but don't actually enable MATC).  I was able to get MATC to work in ElmerGrid a while ago, requiring a few small changes in the code, but couldn't get the cmake side to work properly.  When you get cmake working, it's possible to enable MATC, but that's another effort.

Foadsf:
Thanks — both are good suggestions.

**1. extern "C" in the headers.** Done in 430a0711. You are right that it is simpler, and it turned out better than I expected: `egmain.cpp` no longer needs to change at all, so it is byte identical to `devel` again and drops out of the diff entirely. One file fewer and one file untouched.

I put the guards in the four headers that declare functions — `egconvert.h`, `egmesh.h`, `egnative.h`, `egutils.h` — and left `egdef.h` and `egtypes.h` alone, since those define only macros and structs, which have no linkage, so `extern "C"` there would be a no-op. Say the word if you would rather have them on all six for uniformity.

One thing worth recording: my original wrapper let me prove the linkage boundary by rebuilding the archive from C++-compiled objects and watching the link fail. That check goes vacuous with your approach, because the headers now give C linkage no matter how the translation unit is compiled — which is precisely why yours is the more robust design. I replaced the check with one that strips the guards back out and confirms the link then fails with 50 undefined references, so the guards are demonstrably load-bearing rather than merely present.

**2. MATC in ElmerGrid.** Happy to look at it. From reading the code the CMake side looks tractable: the whole MATC surface in ElmerGrid is `mtc_domath()` and `mtc_init()`, and both are declared inline in `egnative.c` under `#if USE_MATC`, so there is no header to find and it is purely a link question. `matc` is already a real target and `ADD_SUBDIRECTORY(matc)` at `CMakeLists.txt:770` is unconditional and runs before `elmergrid` at :828, so the target always exists in-tree. What blocks it today is that `elmergrid/src/CMakeLists.txt` hardcodes `TARGET_COMPILE_DEFINITIONS(ElmerGrid PRIVATE DISABLE_MATC)`. Replacing that with an option defaulting to OFF, guarded on `TARGET matc` so it degrades cleanly when ElmerGrid is configured standalone, would give the wiring without enabling anything.

I would rather not put that in this pull request, since it is a new capability rather than part of removing the duplicate sources, and this branch is already carrying two borrowed CI commits from #894. I will send it as its own change once these land — unless you would prefer it here.

Foadsf:
Rebased onto `devel` now that #894 has been merged. The two borrowed `ci:` commits are gone, so this pull request is back to just the change it is about.

richb2k:
Changing just the four files by adding Extern C is simpler and sufficient.

Thanks for looking into enabling MATC in ElmerGrid.  Adding the changes later in a separate PR makes sense.

Edit: opened issue #899 to capture this request.

mmuetzel:
These changes look good to me. 👍
Nice reduction of code duplication - a trap that already caught me earlier.

raback:
Thanx Foad for this great development! Initially parts of ElmerGrid where just copy-pasted under ElmerGUI and they started to dangerously differ. Later I took the steps to harmonize them but lacked the skills to actually use the same code under C and C++ projects so hence the dirty script. It is great that we got rid of the redundant code finally.

Issue #202, opened 2020-03-16
cleanup, organization, and modularization


Foadsf:
I took a look at Elmer's [codebase](https://github.com/ElmerCSC/elmerfem), and I should say, to be frank, it is not well organized at all. It seems more of a code dump written and patched over the years without proper organization. Many of the folders don't have README files and the ones which do, sometimes do not offer proper information about the content of that folder. Plain text has also been used instead of modern MarkUp languages such as MarkDown, AsciiDoc, or reStructuredText ...

I think the codebase should be split into several different repositories under the [ElmerCSC](https://github.com/ElmerCSC) GitHub organization:

 - Folders `ElmerGUI`, `ElmerGUIlogger`, and `ElmerGUItester` should be in one repository called ElmerGUI and anything other than the GUI stuff should be scrapped out. I think there are some meshing modules sin the ElmerGUI which hasn't been integrated into the ElmerGrid for example, but I might be wrong on this.
 - The folder `elmergrid` should obviously be an independent repo named ElmerGrid responsible for meshing the input geometries in different formats.
 - The `fem` which is actually the ElmerSolver should be an independent repo called the same name. I think the `fhutiter` folder containing the HUTIter iterative solver should also be placed in the same folder.  
 - The docker related files in `ci` and `docker` folders should be separated just like the HomeBrew which is independent already. 
 - There are two versions of MATC one in root `matc` and the other at `post/matc` where the later is specifically for MVS compilers. This seems bad practice to me. any module should be developed in one single place and we take care of the cross-platform ness in a canonical way. 
 - the `post` folder can be taken out and deprecated. It is important this package is maintained so the people who just start with the software can follow the existing tutorials. we can warn them whenever they open the app that they should use decent alternatives like ParaView instead. 
 - the `elmerice` folder should also be modularized. A great portion of Elmer users have no idea what it is and might never use it, though it is funding wise a big player in the project. 
 - I see that there are third party libraries such as `contrib/Zoltan`, `umfpack`, and even `contrib/lua` embedded into the project. This is in my humble opinion again a bad practice. We should use a decent package manager like Conan, to list all the dependencies and if any of them are not available write Cmake or even bash/batch scripts to download and locate and link against them properly.  
 - we need to have a unified folder structure for any of the above repositories probably following some decent Cmake templates. 
 - Other folders seem to be some snippets or macros for specific purposes (e.g., import from ANSYS) which could be other repose and partly added to he Elmer Documentation. 

[...]

PR #892, opened 2026-08-30
elmergrid: allow configuring as a standalone CMake project

### LINKED ISSUE #891: elmergrid/ cannot be configured as a standalone CMake project
## Summary

`elmergrid/` cannot be configured on its own. It has no `CMAKE_MINIMUM_REQUIRED` or `PROJECT()` of its own, and it depends on three things the Elmer top level project happens to provide. `ReleaseNotes/release_8.3.txt:270` lists "Allow ElmerGrid to be compiled as a standalone application", so I think this is intended to work; at `4f2d7e4b9` on `devel` it does not.

This is also the smallest useful step toward #202, where I suggested ElmerGrid could become its own repository. Whatever is decided there, being able to configure `elmergrid/` by itself is a precondition, and it is worth having on its own since ElmerGrid is frequently used as a standalone mesh conversion tool.

## What is missing

`cmake -S elmergrid -B build` currently fails, because:

1. **No project definition.** `elmergrid/CMakeLists.txt` is three lines: `CONFIGURE_FILE` plus `ADD_SUBDIRECTORY(src)`. There is no `PROJECT()` for CMake to start from.

2. **`-DMINGW32` comes from the root.** `CMakeLists.txt:726` does `ADD_DEFINITIONS(-DMINGW32)` under `IF(MINGW)`. Three ElmerGrid sources need it:

   ```c
   #ifdef MINGW32
     fail = mkdir(directoryname);
   #else
     fail = mkdir(directoryname,0750);
   #endif
   ```

   in `egparallel.c:3961`, `egnative.c:5528` and `egextra.c:966`. Without the definition a mingw build fails with `error: too many arguments to function 'mkdir'; expected 1, have 2`. I hit exactly this. `-DWIN32` from `CMakeLists.txt:730` is in the same position.

3. **`ELMER_FEM_VERSION` comes from the root**, and `config.h.cmake` reads it.

4. **Two include paths are spelled relative to the Elmer roots**, in `elmergrid/src/CMakeLists.txt`:

   ```cmake
   ${CMAKE_SOURCE_DIR}/elmergrid/src/metis-5.1.0/include   # line 37
   ${CMAKE_BINARY_DIR}/elmergrid/src                       # line 42
   ```

   Each names the same directory as the `CMAKE_CURRENT_SOURCE_DIR` / `CMAKE_CURRENT_BINARY_DIR` entry listed beside it when Elmer is the top level project, and a directory that does not exist otherwise. They are redundant in-tree and wrong standalone.

## Suggested fix

Guard a small standalone block on `CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR` so it runs only when `elmergrid/` is the top level project, supplying `PROJECT()`, the `MINGW32`/`WIN32` definitions and a default `ELMER_FEM_VERSION`; and drop the two monorepo-relative include paths in favour of the `CMAKE_CURRENT_*` ones already present. The in-tree build is then untouched, because the guard is false there.

`fempre.c:84-93` already handles a missing branch name and prints `Rev: NA, Compiled: NA`, so a standalone build only needs `ELMER_FEM_VERSION` and can leave the others unset.

I have this working and will open a pull request against `devel`. Both paths build for me with mingw-w64 gcc 15.2 and the two are distinguishable at runtime, the standalone binary reporting `Version: standalone (Rev: NA, Compiled: NA)` and the subdirectory build reporting the version passed down by the parent.

## Environment

`devel` at `4f2d7e4b9`. Reproduced with MSYS2 UCRT64, gcc 15.2.0, CMake with the Ninja generator, on Windows 11. The missing `MINGW32` definition is mingw-specific; the other three points are platform independent.

Foadsf:
Closes #891.

`cmake -S elmergrid -B build` does not work at `4f2d7e4b9`, although `ReleaseNotes/release_8.3.txt:270` lists standalone compilation of ElmerGrid as supported. This makes it work. The evidence for each item is in #891.

## What changed

`elmergrid/CMakeLists.txt` — a standalone block guarded on `CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR`, supplying:

- `CMAKE_MINIMUM_REQUIRED(VERSION 3.12)` (matching the Elmer root) and `PROJECT(ElmerGrid C)`
- `-DMINGW32` / `-DWIN32`, mirroring `CMakeLists.txt:724-731`. Without `MINGW32`, `egparallel.c`, `egnative.c` and `egextra.c` take the two-argument `mkdir()` and a mingw build fails to compile.
- a default `ELMER_FEM_VERSION`. Only that one is needed: `fempre.c:84-93` already handles an unknown branch and prints `Rev: NA, Compiled: NA`.
- a default `CMAKE_BUILD_TYPE` of `Release`

`elmergrid/src/CMakeLists.txt` — drops `${CMAKE_SOURCE_DIR}/elmergrid/src/metis-5.1.0/include` and `${CMAKE_BINARY_DIR}/elmergrid/src`. Each sat next to the `CMAKE_CURRENT_SOURCE_DIR` / `CMAKE_CURRENT_BINARY_DIR` entry naming the same directory when Elmer is the top level project, and named a nonexistent directory otherwise.

Net `+30 / −7`. No source file is touched, and there is no change to what the in-tree build compiles or links.

## Behaviour in the existing build

None. The guard is false whenever `elmergrid` is reached through `ADD_SUBDIRECTORY`, so the added block does not execute, and the two removed include paths were duplicates of entries that remain.

## What I verified

MSYS2 UCRT64, gcc 15.2.0, CMake with Ninja, Windows 11.

**Standalone** — `cmake -S elmergrid -B build-eg-standalone -G Ninja` configures, builds 71/71, links `ElmerGrid.exe`, and the binary runs:

```
ElmerGrid mesh conversion and manipulation utility, Welcome!
Version: standalone (Rev: NA, Compiled: NA)
```

**As a subdirectory** — configured from a small parent project that supplies what the Elmer root supplies (`-DMINGW32`, `-DWIN32`, the four `ELMER_FEM_*` variables, `ELMERSOLVER_RPATH_STRING`) and does `ADD_SUBDIRECTORY(elmergrid)`. Builds 71/71 and the binary reports the parent's values:

```
Version: 9.0-intree-check-devel (Rev: deadbeef, Compiled: 2026-08-30)
```

The two version strings differ, which is the point of running both: it shows the guard is false in the subdirectory case rather than merely that both happen to compile.

Before adding the `MINGW32` definition, the standalone build failed at `egparallel.c:3964`, `egnative.c:5531` and `egextra.c:969` with `too many arguments to function 'mkdir'`, so that part of the change is load-bearing rather than defensive.

## What I could not verify

- **Not a full Elmer build, and no `ctest`.** The subdirectory check above is a stand-in for the Elmer root, not the Elmer root itself; it exercises the interface `elmergrid/` actually uses but does not prove the complete build is unaffected. CI covers that.
- **Only mingw-w64 on Windows.** Not built with MSVC, nor on Linux or macOS. The `MINGW32` point is mingw-specific by nature; the rest is platform independent, but I have not shown it elsewhere.

This is also the smallest concrete step toward #202. Happy to reshape it if you would rather the standalone support were expressed differently.

### DIFF
--- a/elmergrid/CMakeLists.txt
+++ b/elmergrid/CMakeLists.txt
@@ -1,3 +1,37 @@
+# ElmerGrid is configured either as part of the Elmer build, where the top
+# level project supplies the version information, or on its own directly out
+# of this directory.  The block below runs only in the standalone case, so the
+# in-tree build is unaffected.
+#
+# Standalone build:
+#   cmake -S elmergrid -B build-elmergrid && cmake --build build-elmergrid
+IF(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
+  CMAKE_MINIMUM_REQUIRED(VERSION 3.12)
+  PROJECT(ElmerGrid C)
+
+  IF(NOT CMAKE_BUILD_TYPE)
+    SET(CMAKE_BUILD_TYPE "Release" CACHE STRING "Build type" FORCE)
+  ENDIF()
+
+  # The Elmer top level build sets these, and the sources depend on them:
+  # egparallel.c, egnative.c and egextra.c select the one argument form of
+  # mkdir() under MINGW32.  Without the definition the build fails to compile
+  # on a mingw toolchain.  Mirrors CMakeLists.txt in the Elmer root.
+  IF(MINGW)
+    ADD_DEFINITIONS(-DMINGW32)
+  ENDIF()
+  IF(WIN32)
+    ADD_DEFINITIONS(-DWIN32)
+  ENDIF()
+
+  # config.h.cmake normally picks these up from the Elmer top level build.
+  # Only ELMER_FEM_VERSION is set here: fempre.c already handles the case
+  # where the branch is unknown, and prints "Rev: NA, Compiled: NA" then.
+  IF(NOT DEFINED ELMER_FEM_VERSION)
+    SET(ELMER_FEM_VERSION "standalone")
+  ENDIF()
+ENDIF()
+
 CONFIGURE_FILE(config.h.cmake config.h)
 
 ADD_SUBDIRECTORY(src)
--- a/elmergrid/src/CMakeLists.txt
+++ b/elmergrid/src/CMakeLists.txt
@@ -32,14 +32,18 @@ ELSE()
   ADD_SUBDIRECTORY(metis-5.1.0)
 
   # add metis as system include directory
+  # CMAKE_CURRENT_SOURCE_DIR is this directory in both the in-tree and the
+  # standalone build.  The CMAKE_SOURCE_DIR/elmergrid/src spelling that used to
+  # be listed here named the same directory when Elmer is the top level project
+  # and a nonexistent one otherwise.
   TARGET_INCLUDE_DIRECTORIES(ElmerGrid SYSTEM BEFORE PRIVATE
-    ${CMAKE_CURRENT_SOURCE_DIR}/metis-5.1.0/include
-    ${CMAKE_SOURCE_DIR}/elmergrid/src/metis-5.1.0/include)
+    ${CMAKE_CURRENT_SOURCE_DIR}/metis-5.1.0/include)
 ENDIF()
 
+# config.h is generated into the elmergrid binary directory and reached as
+# "../config.h" from the sources here, so this directory is what is needed.
 TARGET_INCLUDE_DIRECTORIES(ElmerGrid PRIVATE
-  ${CMAKE_CURRENT_BINARY_DIR}
-  ${CMAKE_BINARY_DIR}/elmergrid/src)
+  ${CMAKE_CURRENT_BINARY_DIR})
 
 TARGET_COMPILE_DEFINITIONS(ElmerGrid PRIVATE DISABLE_MATC)
 


Foadsf:
The failing `macos-15-intel (using openblas)` check here is not caused by this pull request. It fails in the `install dependencies` step, before anything is configured or compiled, with `vtk: no bottle available!` — Homebrew publishes no Intel macOS bottle for vtk (as of 9.7.0 the tags are arm64_tahoe, arm64_sequoia, arm64_sonoma, arm64_linux, x86_64_linux).

The same job and step fail on unrelated branches, including `fix/permafrost-hydraulic-reference-state` (run 33298600360) several hours before this branch existed, and `interfrost-devel` the day before. The last green runs of that workflow were on 2026-08-28.

Filed as #893 with a fix in #894. The `macos-14` legs pass here.

Foadsf:
I have temporarily cherry-picked the two commits from #894 onto this branch so that CI can actually run here. Without them the `macos-15-intel` leg dies in `install dependencies` on `vtk: no bottle available!` before anything in this pull request is compiled, as described above.

Those two `ci:` commits are **not part of this change** and are identical to the ones in #894 (same patches, same resulting workflow file). Once #894 is merged I will rebase this branch onto `devel`, which drops them automatically, leaving only the commit this pull request is actually about.

If you would rather I keep this branch free of them and simply leave the macOS check red until #894 lands, say so and I will drop them again.

Foadsf:
Rebased onto `devel` now that #894 has been merged. The two borrowed `ci:` commits are gone, so this pull request is back to just the change it is about.

raback:
Nice feature! There may be some people using ElmerGrid just for file conversion.

PR #898, opened 2026-08-31
CI (macos): Switch to newer versions of macOS on Apple Silicon.

mmuetzel:
Homebrew started (early?) to move support for Intel processors to their "Tier 3".
See: https://docs.brew.sh/Support-Tiers#future-macos-support
The `vtk` package is already no longer available as a binary package for macOS 15 on Intel.

Move the corresponding workflow matrix element to Apple Silicon runners. Also move the remaining workflow matrix elements to an image with a newer version of macOS.

More and more packages for Intel hardware are likely to disappear from Homebrew. So, I think this is a better alternative in the long run to #894.

### DIFF
--- a/.github/workflows/build-macos-homebrew.yaml
+++ b/.github/workflows/build-macos-homebrew.yaml
@@ -25,24 +25,11 @@ jobs:
       fail-fast: false
 
       matrix:
-        os: [macos-14]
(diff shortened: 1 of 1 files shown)


juharu:
Thanks Markus!

PR #908, opened 2026-09-03
FindARPACK/FindPARPACK: search for the library before consulting the config package

### LINKED ISSUE #907: FindARPACK/FindPARPACK: a broken distribution config package takes the configure down, and the result is tested with IF(ARPACK_FOUND) instead of IF(TARGET arpack)
`FindARPACK.cmake` consults the distribution's CMake config package with `FIND_PACKAGE(ARPACK CONFIG ...)`, and on two installations in wide use that call raises `FATAL_ERROR` while it is being executed, taking the whole configure down. `FindPARPACK.cmake` behaves the same way.

The two installations fail for different reasons, which matters for any fix:

**Debian and Ubuntu.** `arpackng-config.cmake` does `include(...arpackngTargets.cmake)` and that file is not packaged, so the `include()` itself fails:

```
CMake Error at /usr/lib/x86_64-linux-gnu/cmake/arpack-ng/arpackng-config.cmake:35 (include):
  include could not find requested file:
CMake Error at cmake/Modules/FindARPACK.cmake:28 (GET_TARGET_PROPERTY):
CMake Error at cmake/Modules/FindARPACK.cmake:29 (GET_TARGET_PROPERTY):
CMake Error at cmake/Modules/FindARPACK.cmake:32 (GET_TARGET_PROPERTY):
```

**MSYS2, when `arpack` is installed without `parpack`.** Here every config file *is* present — @mmuetzel is right that the package is complete. The problem is that `arpackngTargets.cmake` is shipped by the `arpack` package and declares an imported `parpack` target unconditionally:

```
$ pacman -Qo /ucrt64/lib/cmake/arpackng/arpackngTargets.cmake
... is owned by mingw-w64-ucrt-x86_64-arpack 3.9.1-6

$ grep -n parpack /ucrt64/lib/cmake/arpackng/arpackngTargets.cmake
22:foreach(_cmake_target IN LISTS _cmake_import_check_targets)   # arpack parpack
66:# Create imported target parpack
67:add_library(parpack SHARED IMPORTED)

$ pacman -Qo /ucrt64/lib/libparpack.dll.a
... is owned by mingw-w64-ucrt-x86_64-parpack 3.9.1-6
```

The library is in a *different* package. Its own `_cmake_import_check_files_for_parpack` loop then raises:

```
CMake Error at C:/msys64/ucrt64/lib/cmake/arpackng/arpackngTargets.cmake:91 (message):
  The imported target "parpack" references the file
     "C:/msys64/ucrt64/lib/libparpack.dll.a"
  but this file does not exist.
```

Reproduced with a project that does nothing but `FIND_PACKAGE(ARPACK REQUIRED)` against Elmer's module: it exits 1 with `arpack` installed and `parpack` not, and exits 0 once `parpack` is installed. So MSYS2 is not broken as a platform; that particular combination is, and the two packages are independently installable.

### The second problem, which is Elmer's own

After the `FIND_PACKAGE(... CONFIG)` call, `FindARPACK.cmake` tests `IF(ARPACK_FOUND)`. CMake sets that variable because a config file was located and ran to its end, **even when it created no target**, so the three `GET_TARGET_PROPERTY` calls that follow are made against a target that does not exist and contribute three further errors of their own. `IF(TARGET arpack)` is the honest test.

### Why file-existence checks are not enough

The obvious guard — check that the targets file the config `include()`s actually exists — catches Debian and Ubuntu but not MSYS2, where all four config files are present and the failure comes from a consistency check *inside* one of them. Distinguishing a sound installation from a broken one appears to require actually executing the config, which is the thing that cannot be done safely in-process.

### Why it matters now

@juharu hit this on Ubuntu while testing whether ARPACK-NG could replace the bundled sources, and reported in #844 that he "needed to give absolute paths of the libraries". The diagnosis above is his; this is filed separately because it is a defect in a shared Find module that stands on its own, whatever is decided about the bundled copy.

It matters more if `EXTERNAL_ARPACK=ON` ever becomes the default, since then every build on an affected installation meets it.

PR #908.

Foadsf:
Fixes #907.

`FIND_PACKAGE(... CONFIG)` executes the config file it finds, so a distribution shipping a broken one makes `FindARPACK.cmake` fail in a way that testing the result afterwards cannot prevent. Debian and Ubuntu install an `arpackng-config.cmake` whose `arpackngTargets.cmake` is not packaged; MSYS2 ships one declaring a `parpack` target from a separate package that need not be installed. Either aborts the configure while the file is being read, and the manual search below — which finds these libraries without trouble — is never reached.

Two changes, both small:

1. Run the plain `FIND_PATH`/`FIND_LIBRARY` search first, and consult the config package only if it finds nothing. That costs nothing when the config package is good and avoids opening it when it is not.
2. Where the config package is still used, test `IF(TARGET arpack)` rather than `IF(ARPACK_FOUND)`. CMake sets `ARPACK_FOUND` because a config file was located and ran to its end, even when it created no target, so the three `GET_TARGET_PROPERTY` calls that followed landed on a target that did not exist and added three errors of their own.

`FindPARPACK.cmake` gets the same two, plus one specific to it: it now accepts `ARPACK_INCLUDE_DIR` when no `parpack.h` is found under its own name. arpack and parpack are a single package on some distributions and share an include directory on others, so without that a working `libparpack` still fell through into the config package looking for a header that was never going to be there.

### Measured

MSYS2 UCRT64, `mingw-w64-ucrt-x86_64-arpack` installed and `mingw-w64-ucrt-x86_64-parpack` deliberately not, against a project that does nothing but `FIND_PACKAGE(ARPACK REQUIRED)` with these modules on `CMAKE_MODULE_PATH`:

```
########## before ##########
exit: 1
CMake Error at C:/msys64/ucrt64/lib/cmake/arpackng/arpackngTargets.cmake:91 (message):
########## after ##########
exit: 0
-- PROBE_RESULT lib=C:/msys64/ucrt64/lib/libarpack.dll.a inc=C:/msys64/ucrt64/include/arpack
```

With `parpack` present both succeed, so nothing that worked before stops working. The reordered version reports the import library rather than the DLL, which is what the rest of Elmer's `FIND_LIBRARY` calls report as well.

The Ubuntu half of #907 is from CI logs rather than from my own machine, and the diagnosis in that issue is @juharu's, from #844.

### Not fixed

No ARPACK anywhere the plain search looks, plus a broken config present: the config still gets opened and still errors. Avoiding that would mean running `find_package` in a separate CMake process, which seemed a lot of machinery for a system with no usable ARPACK on it.

### Scope

This is independent of what is decided about the bundled ARPACK in #844. It matters either way, and rather more if `EXTERNAL_ARPACK=ON` ever becomes the default, since then every build on these platforms meets it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cmake/Modules/FindARPACK.cmake
+++ b/cmake/Modules/FindARPACK.cmake
@@ -22,45 +22,61 @@ ENDIF()
 
 IF(NOT ARPACK_FOUND)
   MESSAGE(STATUS "Finding arpack libraries")
-  # Try to find with CMake config file of upstream arpack.
-  FIND_PACKAGE(ARPACK CONFIG NAMES arpack arpackng arpack-ng)
-  IF(ARPACK_FOUND)
-    GET_TARGET_PROPERTY(ARPACK_INCLUDE_DIR arpack INTERFACE_INCLUDE_DIRECTORIES)
-    GET_TARGET_PROPERTY(ARPACK_LIBRARIES arpack IMPORTED_LOCATION_RELEASE)
-    # Check if a debug build type was used
-    IF(NOT ARPACK_LIBRARIES)
-      GET_TARGET_PROPERTY(ARPACK_LIBRARIES arpack IMPORTED_LOCATION_DEBUG)
-    ENDIF()
-  ENDIF()
 
-  # There is no arpack-config script or something went wrong with the script
-  # Fall back to manual search
+  # Plain search first; the upstream CMake config package is consulted only if
+  # it finds nothing.
+  #
+  # FIND_PACKAGE(... CONFIG) does not merely look for a config file, it
+  # executes it, so a distribution shipping a broken one makes this module fail
+  # in a way no test of the result can prevent. Two in circulation today:
+  # Debian and Ubuntu install an arpackng-config.cmake whose
+  # arpackngTargets.cmake is not packaged, and MSYS2 ships one that declares a
+  # parpack target from a separate package that need not be installed. Both
+  # abort the configure while reading the file.
+  #
+  # Searching for the library directly costs nothing when the config package is
+  # good, and avoids opening it at all when it is not.
+  INCLUDE(FindPackageHandleStandardArgs)
+  MESSAGE(STATUS "Searching for arpack library")
+
+  # Try to find ARPACK header
+  SET(ARPACKINCLUDE
+    "${ARPACK_ROOT}/include"
+    "$ENV{ARPACK_ROOT}/include"
+    "${ARPACKROOT}/include"
+    "$ENV{ARPACKROOT}/include"
+    INTERNAL
+    )
+  FIND_PATH(ARPACK_INCLUDE_DIR NAMES arpack.h arpackng.h arpack-ng.h
+    HINTS ${ARPACKINCLUDE} PATH_SUFFIXES arpack arpackng arpack-ng)
+
+  # Try to find ARPACK libraries
+  SET(ARPACKLIB
+    "${ARPACK_ROOT}/lib"
+    "$ENV{ARPACK_ROOT}/lib64"
+    "${ARPACKROOT}/lib"
+    "$ENV{ARPACKROOT}/lib64"
+    INTERNAL
+    )
+  FIND_LIBRARY(ARPACK_LIBRARIES NAMES arpack arpackng arpack-ng HINTS ${ARPACKLIB})
+
+  # Nothing found by hand: try the config package after all.
   IF(NOT ARPACK_LIBRARIES OR NOT ARPACK_INCLUDE_DIR)
-    # Fall back to manual search
-    INCLUDE(FindPackageHandleStandardArgs)
-    MESSAGE(STATUS "Searching for arpack library")
-
-    # Try to find ARPACK header
-    SET(ARPACKINCLUDE
-      "${ARPACK_ROOT}/include"
-      "$ENV{ARPACK_ROOT}/include"
-      "${ARPACKROOT}/include"
-      "$ENV{ARPACKROOT}/include"
-      INTERNAL
-      )
-    FIND_PATH(ARPACK_INCLUDE_DIR NAMES arpack.h arpackng.h arpack-ng.h
-      HINTS ${ARPACKINCLUDE} PATH_SUFFIXES arpack arpackng arpack-ng)
-
-    # Try to find ARPACK libraries
-    SET(ARPACKLIB
-      "${ARPACK_ROOT}/lib"
-      "$ENV{ARPACK_ROOT}/lib64"
-      "${ARPACKROOT}/lib"
-      "$ENV{ARPACKROOT}/lib64"
-      INTERNAL
-      )
-    FIND_LIBRARY(ARPACK_LIBRARIES NAMES arpack arpackng arpack-ng HINTS ${ARPACKLIB})
-
+    # Try to find with CMake config file of upstream arpack.
+    FIND_PACKAGE(ARPACK CONFIG NAMES arpack arpackng arpack-ng)
+    # IF(TARGET arpack) rather than IF(ARPACK_FOUND): CMake sets ARPACK_FOUND
+    # when a config file was located and ran to its end, even if it created no
+    # target, and GET_TARGET_PROPERTY on a target that does not exist adds
+    # three more errors to whatever the config file already reported. The
+    # target is what is actually wanted here.
+    IF(TARGET arpack)
+      GET_TARGET_PROPERTY(ARPACK_INCLUDE_DIR arpack INTERFACE_INCLUDE_DIRECTORIES)
+      GET_TARGET_PROPERTY(ARPACK_LIBRARIES arpack IMPORTED_LOCATION_RELEASE)
+      # Check if a debug build type was used
+      IF(NOT ARPACK_LIBRARIES)
+        GET_TARGET_PROPERTY(ARPACK_LIBRARIES arpack IMPORTED_LOCATION_DEBUG)
+      ENDIF()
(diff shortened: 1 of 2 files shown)


Foadsf:
Both MSYS2 external-dependency jobs failed on the first push, and the cause was worth having: the reorder was fine, the imported target was not.

```
-- A library with ARPACK API found.
-- ARPACK libraries: D:/a/_temp/msys64/ucrt64/lib/libarpack.dll.a
...
ninja: error: 'arpack-NOTFOUND', needed by 'fem/src/libelmersolver.dll'
```

ARPACK was found; the variable was right. What broke was the target. When the manual search succeeds, this module defines the imported target itself, and it declared it `SHARED` with only `IMPORTED_LOCATION` set. A `SHARED` imported target also needs `IMPORTED_IMPLIB` on Windows, so anything linking the bare target name — which `fem/src/CMakeLists.txt` does, through `arpack` in `ELMERSOLVER_LIBRARIES` — resolved to `arpack-NOTFOUND`.

That defect predates this branch. Reordering the search is simply what made Windows reach that code path for the first time, because the config package used to create the target instead. `ADD_LIBRARY(... UNKNOWN IMPORTED)` links `IMPORTED_LOCATION` directly and is right whether `FIND_LIBRARY` returned a `.so`, a `.a` or an import library. Same change in `FindPARPACK.cmake`.

My original probe only configured, which is why I did not catch this before opening the PR. It now builds an executable that links the bare target, which is what Elmer actually does. On MSYS2 UCRT64:

| | configure | link |
|---|---|---|
| devel | 0 | 0 &nbsp;(target came from the config package) |
| reorder alone | 0 | 1 &nbsp;`ninja: 'arpack-NOTFOUND'` |
| reorder + `UNKNOWN` | 0 | 0 &nbsp;`type=UNKNOWN_LIBRARY` |

and with `parpack` deliberately uninstalled, where the config package cannot be read at all:

| | configure | link |
|---|---|---|
| devel | 1 &nbsp;`CMake Error at arpackngTargets.cmake:91` | — |
| this PR | 0 | 0 |

mmuetzel:
I'm not sure this goes in a "good" direction. See my comment in #907.

That concern actually realized in your test: Some additional information (that would have been provided by the imported target) was missing when it was ignored.

Foadsf:
You are right, and the test proved your point rather than mine — thank you for catching it before this landed.

`arpack-NOTFOUND` was precisely the imported target's information going missing. Once the config package was skipped, the module synthesised its own target from a bare `FIND_LIBRARY` path, and that target lacked what the real one carried. I patched the symptom by declaring it `UNKNOWN` instead of `SHARED`; you are pointing at the cause, which is that the target should not have been discarded at all. Preprocessor definitions and transitive link dependencies would have gone the same way with nothing as visible as a link error to reveal it.

So I have reversed it, along the lines you suggested: the config package is consulted **first** and its target is used exactly as before; the plain search is reached only when the config is genuinely unusable.

The detection is the awkward part. Checking that the file the config `include()`s exists catches Debian and Ubuntu, but not MSYS2, where all four config files are present and the failure comes from a consistency check *inside* `arpackngTargets.cmake`. The only thing that distinguishes a sound installation from a broken one in both cases is executing the config, and a `FATAL_ERROR` cannot be caught in-process. So it is executed in a throwaway CMake process first — same generator, compiler and prefix path — and loaded here only if that process survives. The verdict is cached, so it costs one configure of a trivial project, once.

Measured on MSYS2 UCRT64, against a project that does `FIND_PACKAGE(ARPACK REQUIRED)` and then links the bare target:

| | `devel` | this PR |
|---|---|---|
| `arpack` + `parpack` installed | configure 0, link 0, target `SHARED_LIBRARY` from the config | configure 0, link 0, target `SHARED_LIBRARY` **from the config** |
| `arpack` only | configure **1**, `arpackngTargets.cmake:91` | configure 0, link 0, target `UNKNOWN_LIBRARY` from the plain search |

The first row is the one that matters for your objection: where the installation is sound, nothing changes. Same target, same provenance, same reported library. The fallback only appears in the row where `devel` cannot configure at all.

I have also corrected #907. The title claimed MSYS2 was affected, full stop, which is not right — you were correct that the package contains everything. The failure needs `arpack` installed without `parpack`, because `arpackngTargets.cmake` ships in the `arpack` package and declares an imported `parpack` target whose library is in the separate `parpack` package. The issue now says that, with the `pacman -Qo` output showing which package owns which file.

If you would rather have the file-existence heuristic than a spawned CMake process, say so and I will send that instead — it is cheaper and covers Debian, at the cost of leaving the MSYS2 case unfixed.

mmuetzel:
Sorry. But again, I'm not sure whether this is the right approach.
Rather than requiring that each project that tries to import the ARPACK-ng CMake targets would need to jump through these hoops, this should be fixed upstream.
I can try to look at the MSYS2 part over the weekend. (That is something that the packager, i.e. MSYS2, should improve.)

Regarding the issue in Debian, I left a comment at https://github.com/opencollab/arpack-ng/pull/438 which introduced this issue in upstream ARPACK-ng.
If a solution for this issue materializes upstream, it should probably also be reported to Debian (and Ubuntu).
I'm not using Debian myself. So, someone else should report the issue to them when a solution was accepted upstream.

mmuetzel:
I've opened https://github.com/msys2/MINGW-packages/pull/31399 that addresses this issue in the packages from MSYS2.

mmuetzel:
See also #910 which should address a related issue that I stumbled upon while testing with different configurations of ARPACK-ng and different configurations of Elmer.

mmuetzel:
The issue with the CMake config files of ARPACK in MSYS2 should be fixed after updating to the latest package versions (e.g., with `pacman -Syu`).

PR #918, opened 2026-09-11
Relocate prefix of `defaults.lua` file on runtime

mmuetzel:
The prefix to the `defaults.lua` file is currently hard-coded using the preprocessor macro `ELMER_SOLVER_HOME`.
It is possible to override that by setting the environment variable `ELMER_HOME`. But the prefix is also available on runtime by using the function `GetSolverHome`.

Keep the existing two sources for the prefix in the established priority. But add a fallback to `GetSolverHome` if the file could not be located with the other two.

The used logic is very much inspired by how this is done for the file `SOLVER.KEYWORDS` in the same source file.

This should also be addressing an old TODO note in that block from 2018.


mmuetzel:
I'm unsure if it would be better to drop relying on the preprocessor macro `ELMER_SOLVER_HOME` entirely.
What do you think?

king-aj3:
Tested this PR's Windows artifact (`elmerfem-nompi-918_merge-26.2-bb658616a-20260911_Windows-AMD64.zip`) against the devel artifact from the day before (`8ae4a05b8`), Windows 11, non-build machine, same subprocess method as the wrapper tests. This time the subject is `ElmerSolver` itself, on a real case: the #860 live check (4×4-quad unit square from a `.grd` via `ElmerGrid`, steady HeatSolve, `Heat Conductivity` from a UDF compiled with the artifact's own `elmerf90`) plus two things that only work if `defaults.lua` was actually loaded — a `!---LUA BEGIN … END` block (extracted by `readsif`, which `defaults.lua` defines) and `Heat Source = Real #pi*src_scale` (`pi` is an alias from `defaults.lua`, `src_scale` comes from the block). `<EH>\bin` on PATH throughout; `ElmerGrid` and `elmerf90` ran with no `ELMER_*` variable on both artifacts.

| case | environment | devel `8ae4a05b8` (pre-PR) | PR #918 `bb658616a` |
|---|---|---|---|
| S1 | `ELMER_HOME=<EH>` | exit 0, `ALL DONE`, UDF loaded, norm 0.21340504 | same, norm 0.21340504 |
| S2 | **no `ELMER_*` variable at all** | exit 1 — `FATAL Lua error: [string "loadfile("D:/a/elmerfem/elmerfem/usr/share/..."]:1: attempt to call a nil value` | **exit 0, `ALL DONE`, UDF loaded, norm 0.21340504** |
| S3 | `ELMER_HOME=C:\old_elmer_that_moved` (stale) | exit 1 — `InitializeElementDescriptions: elements.def not found` | exit 1 — same |
| S4 | `ELMER_HOME=<EH>` + `ELMER_LIB=<EH>\share\elmersolver\lib` | exit 0, norm 0.21340504 | same |

**S2 is the point of the PR and it works**: with nothing set, the solver finds `defaults.lua` relative to its own executable, the Lua block and `#pi` evaluate, the UDF loads, and the result is bit-identical to the `ELMER_HOME` run. For context, on the current zips `ElmerSolver` cannot run at all without `ELMER_HOME` — `loadfile()` on the baked CI prefix returns nil and calling it is fatal — so this is not a convenience, it is the difference between the zip working out of the box and not.

**Two things I measured on the way, neither a regression:**

1. **MAX_PATH bites the new lookup before the old ones.** My first run of S2 on this PR failed with the new `LoadModel: defaults.lua not found` — at an unpack location 211 characters deep. `GetSolverHome` returns the exe-relative path un-normalized (`<EH>\bin\..\share\elmersolver`, 236 chars here); with `/lib/elements.def` appended that is 253 characters and `INQUIRE` finds it, with `/lua-scripts/defaults.lua` it is **261**, and the ANSI file APIs behind `INQUIRE` stop at 260. Same trees reached through a 32-character junction: everything green (the table above). The zip's own directory name is 66 characters, so deep OneDrive-style paths will reach this. Normalizing away the `\bin\..` in `getsolverhome` (Load.c:131) buys 7 characters cheaply; the hard limit is `INQUIRE` itself unless long-path-aware calls are used.

2. **A stale `ELMER_HOME` still kills the solver at `elements.def` (S3), and the new fallback chain cannot help**, because the third source is not independent of the first: `getsolverhome` (Load.c:99-111) returns `%ELMER_HOME%\share\elmersolver` verbatim when the variable is set, with no existence check, so both `InitializeElementDescriptions` and the new `defaults.lua` block try the same wrong prefix twice and never reach the exe-relative path. The `INQUIRE`-then-fall-through pattern this PR uses would close it if `getsolverhome` also fell through to the exe location when the `ELMER_HOME` candidate does not exist. (The user who hits this is the one who set `ELMER_HOME` for an old install and unpacked a new zip elsewhere.)

On the question of dropping the `ELMER_SOLVER_HOME` macro: from the Windows zip's side it has never pointed anywhere valid on a user's machine — the `D:/a/elmerfem/...` path in the S2 failure above is that macro — so nothing a zip user can do depends on it. I have no view on how Linux distribution builds rely on it.

Happy to rerun on a later artifact.

mmuetzel:
You are raising good points.

Following the principle of "the user is always right", we should never use any fallback values if `ELMER_HOME` has been set by the user.
The original implementation did that. And we should probably keep it that way.
If `ELMER_HOME` is set incorrectly, the original implementation crashes. I think we could improve on that though.

I'll amend the PR to do that.

Regarding the path length limit, the original implementation potentially cut of the LUA command. That could have led to a syntax error in that command (e.g., no closing quotes or parenthesis).
I was interpreting the comment to refer to that.
That should no longer happen with the changes. 
We could still be hitting a Windows API limit like you wrote though.
And, the fixed length buffer that we need to share a string in a parallel section could also be too short.

I'll at least add a comment about that limit if there is no better solution.

I was wondering about the parallel section anyway. If all of it is also "critical", does anything run multi-threaded at all?

mmuetzel:
I amended the change to respect what the user set for `ELMER_HOME` (even if it might be invalid). If it isn't valid, it should terminate the program with a fatal error instead of crashing.

A follow-up change should help avoiding the path length limit in that part of the code. 
The overall supported path length might still be limited by `MAX_PATH_LEN` (which is set to 4096 currently) + `n` (the length of the appended part of the path to `defaults.lua`), the Fortran runtime implementation (`GET_ENVIRONMENT_VARIABLE`), the used Windows API (`GetModuleFileNameA`), or limits in the implementation of Lua.

I also took the opportunity to add a couple of comments why the PARALLEL OpenMP block looks like it does. (It took me a bit to figure that out.)

king-aj3:
Retested the amended PR (`elmerfem-nompi-918_merge-26.2-a28897559-20260912_Windows-AMD64.zip`), same HeatSolve + UDF + Lua-block + `#pi` case, two locations.

| case | environment | 32-char junction | 211-char unpack path |
|---|---|---|---|
| S1 | `ELMER_HOME=<EH>` | exit 0, norm 0.21340504 | exit 0, norm 0.21340504 |
| S2 | no `ELMER_*` at all | **exit 0, `ALL DONE`, UDF loaded, norm 0.21340504** | exit 1 — `LoadModel: defaults.lua not found at <EH>\bin\..\share\elmersolver/lua-scripts/defaults.lua` (that string is 261 characters) |
| S3 | `ELMER_HOME` stale | exit 1 — `InitializeElementDescriptions: elements.def not found` | same |
| S4 | `ELMER_HOME` + `ELMER_LIB=<EH>\share\elmersolver\lib` | exit 0, norm 0.21340504 | same |
| S5 | `ELMER_LIB` valid + `ELMER_HOME=C:\old_elmer_that_moved` | exit 1 — **`LoadModel: defaults.lua not found at C:\old_elmer_that_moved/share/elmersolver/lua-scripts/defaults.lua`** | same |

S5 is the new case: `elements.def` is found through `ELMER_LIB`, so the solver reaches the Lua init with a stale `ELMER_HOME`, and the amended code does what you described — respects the user's setting and terminates with a message that shows exactly where it looked, stale value included. That is all the user needs to fix it. S3 still dies earlier in `InitializeElementDescriptions`, which this PR does not touch.

The 211-char column is the MAX_PATH point from before, unchanged as expected; the new message now prints the offending un-normalized path itself, which is a better diagnostic than the old silent nil call in any case. Thanks for the OpenMP comments — and the `elmer_home` FIXME is the right note to leave for the follow-up.

mmuetzel:
The place where the length of the file path is limited is likely the following in Lua itself:
https://github.com/lua/lua/blob/7579fc9d7ed90240487251dfb69168f8e64e9294/lauxlib.c#L834
```
    lf.f = fopen(filename, "r");
```
Or here in Elmer's bundled version of Lua:
https://github.com/ElmerCSC/elmerfem/blob/17e321161179de25c11603ab428d55268f36eada/contrib/lua-5.1.5/src/lauxlib.c#L564

Afaict, `fopen` on Windows has a path length limit of 260 characters by default. It might be possible to circumvent that limit with some registry changes. But that might cause issues for other programs if I'm not mistaken.

Additionally, that also restricts the support to characters in the current (single-byte) code page on Windows.

In any case, this would likely need changes in Lua. E.g., it might be possible for them to add an OS-specific implementation (of parts) of their `loadfile` function. 
Supporting the whole Unicode character set on Windows would probably be harder though and might require more changes all over Lua.
There is probably little that can be done about that in Elmer itself at the time being.

mmuetzel:
I amended with a change that adds a more fitting "FIXME" note about the current file path length limit.
No functional changes to the previous state (other than rebasing on the current head of the `devel` branch).

king-aj3:
Rebased head (`e79e3e17d`, artifact `20ad1a146`) retested: identical to the amended run — the no-`ELMER_*` case exits 0 with the UDF loaded and norm 0.21340504 at a short path, the stale-`ELMER_HOME` case reports the exact path it tried, and the deep-path case fails at the 260 limit exactly where your `fopen` analysis says it must. Agreed that this one belongs to Lua, not Elmer — the FIXME is the right place to leave it.

mmuetzel:
Amended the PR to make the new subroutine `InitializeLua` internal to `LoadModel`.

mmuetzel:
I removed the `CRITICAL` block locally. I can't see any issues for multiple different values of `OMP_NUM_THREADS` or parallel jobs in the following:
```
OMP_NUM_THREADS=32 ctest -L lua -j 16 .
```
I intentionally pushed the number of parallel threads to something unreasonable for a processor with 8 cores (16 logical).
I know that it might be difficult to reproduce potential concurrency issues. But are we sure Lua isn't actually thread-safe (and the `CRITICAL` wasn't added for something that has been fixed in the meantime)?

Should I amend this with the following? Or is there something obvious that requires running the initialization serially?
```
 fem/src/ModelDescription.F90 | 4 ----
 1 file changed, 4 deletions(-)

diff --git a/fem/src/ModelDescription.F90 b/fem/src/ModelDescription.F90
index 96cd2419b..01216b39c 100644
--- a/fem/src/ModelDescription.F90
+++ b/fem/src/ModelDescription.F90
@@ -3502,9 +3502,6 @@ CONTAINS
       !$OMP          Private(txcmd, ompthread, lstat, loadfile_cmd) &
       !$OMP          Default(none)
 
-      ! The Lua functions are not thread-safe.
-      ! Run initialization for each thread serially.
-      !$OMP CRITICAL
       LuaState = lua_init()
       IF(.NOT. LuaState % Initialized) THEN
         CALL Fatal(Caller, 'Failed to initialize Lua subsystem.')
@@ -3529,7 +3526,6 @@ CONTAINS
       lstat = lua_dostring(LuaState, 'loadstring(readsif("'//trim(ModelName)//'"))()' // c_null_char)
       lstat = lua_dostring(LuaState,  trim(txcmd)// c_null_char)
       LuaState % tx => lua_getusertable(LuaState, 'tx'//c_null_char)
-      !$OMP END CRITICAL
       !$OMP END PARALLEL
     END SUBROUTINE InitializeLua
 #endif
```

king-aj3:
Before the `CRITICAL` goes: one thing to weigh against the clean `ctest -L lua` run. In #925 @juharu traced a `beamer3d_box2` flake to a deferred-length allocatable character used from parallel threads ("not thread safe — essentially broken — in gfortran"). `InitializeLua` has one — `CHARACTER(LEN=:), ALLOCATABLE :: loadfile_cmd`, `Private` in the parallel region. Serialized by the `CRITICAL` that is fine; without it, it is the same construct. A fixed-length buffer there would sidestep the question regardless of what Lua itself tolerates.

(This is from reading the two threads side by side, not from a measurement.) Happy to run the HeatSolve + Lua case at `OMP_NUM_THREADS` 1/4/8/16 on the Windows artifact once the change is pushed — MinGW's threadprivate implementation is not what the Linux runners exercise.

mmuetzel:
I'm not an expert when it comes to OpenMP with GFortran. But to me it looked like it didn't actually provide pointers to valid memory when an allocatable array from a surrounding context was directly used in a parallel OpenMP construct. That was reproducible 100% of the time (with or without `CRITICAL`). It doesn't seem to be a concurrency issue but a general limitation (of the OpenMP implementation in GFortran - I'd guess).

My interpretation is that @juharu was referring to that issue.

The indirection through a subroutine call seems to have "sanitized" that issue here. The variable-length argument actually holds the expected content in all threads.

Edit: Essentially, the same is the case for the `ModelName` parameter. (Even if that comes through more than one subroutine indirection, it is "originally" an deferred length allocatable.)

mmuetzel:
Maybe, we should leave the question about `CRITICAL` to a different issue or pull request.
It isn't actually related to the relocation issue this PR tries to fix. It just happens to be in the area of the source code that this PR changes.

juharu:
The problem is not the character string in itself -  but from reading the object dumps -  the
"length" of the deferred-lenght-allocatable-character is shared. The breakup would require
the lengths of the strings to differ - if they are equal it should be ok, at least it'd be a very
rare occurance to have troubles.

The length trouble has been verified several times also in practice.

mmuetzel:
@juharu: The length of the string is the same for all threads. Actually, the string itself is the same.

Leaving the question about `CRITICAL` aside: Do the changes in this PR look ok to you?

juharu:
Right, the changes seem OK to me. Probably also safe as the function returning the allocatable character string in itself is not of the same type. I'd keep the CRITICAL anyway for safety.

juharu:
Btw, from reading the code: is there a reason why  "loadfile_cmd" string  is set (as opposed to just read) in the parallel section at all (private -> shared) ? (also referring the "default_lua" there then seems unnecessary).  I have become a bit paranoid 'bout the parallel sections recently, I guess...

juhanikataja:
This looks good idea, but it should give warnings (or even error!) if ELMER_HOME is defined and defaults.lua is not found under that tree and not silently move on to next location to read it from.

> Btw, from reading the code: is there a reason why "loadfile_cmd" string is set (as opposed to just read) in the parallel section at all (private -> shared) ? (also referring the "default_lua" there then seems unnecessary). I have become a bit paranoid 'bout the parallel sections recently, I guess...

I don't think there is a reason.

mmuetzel:
Good points.

I clarified the comments to (hopefully) better describe the logic:
If `ELMER_HOME` is set, it is used unconditionally. There is no fallback to the preprocessor macro `ELMER_SOLVER_HOME` or the `GetSolverHome` function.
If it is set to something with which the `defaults.lua` file cannot be found, an error is emitted (including the path where it looked for that file):
```
      IF (.NOT. fexist) THEN
        CALL Fatal(Caller, 'defaults.lua not found at ' // defaults_lua)
      END IF
```
Only if `ELMER_HOME` is unset or empty, it uses the preprocessor macro `ELMER_SOLVER_HOME` or the `GetSolverHome` function.

The only reason why `defaults_lua` was used as the function argument for the new internal subroutine `InitializeLua` is that the PR "evolved that way".
I refactored it to pass `loadfile_cmd` as the argument instead.

king-aj3:
Amended head (`e7dbb9ed2`, artifact `674211ca8`) retested on Windows: identical to the previous two runs — the no-`ELMER_*` case exits 0 with the UDF loaded and norm 0.21340504, `ELMER_HOME` set-but-wrong terminates naming the path it tried, and the deep-path case still stops at the 260-character limit as expected. Moving `loadfile_cmd` out of the parallel region changed nothing observable, which is what a refactor should do.

mmuetzel:
Thanks for testing again.

Do the changes look good to you now? Ok to merge?

juharu:
Thanks!

PR #912, opened 2026-09-08
Support relocating the installation prefix on runtime for the compiler wrappers on Windows

mmuetzel:
Currently users must set environment variables like `ELMER_HOME` to be able to use Elmer on Windows.

It should be possible to infer that information from the location where executables or libraries are installed.
Do that for the compiler wrappers (as a first step).

If that works out satisfactory, it might be possible to use the same strategy to relocate the installation prefix on runtime also in other places which currently rely on the value of environment variables.
Maybe, it would even be possible to use Elmer without the need to set any environment variable.

PR #921, opened 2026-09-12
Prepend path of compiler wrapper to PATH before spawning child on Windows

mmuetzel:
Windows does not have an RPATH mechanism. Instead, the OS resolves DLL dependencies on load time by checking the working directory and the directories in the PATH environment variable.

Prepend the PATH environment variable with the path to the wrapper executable (assuming that all DLLs for child and grandchild processes are in that folder).


king-aj3:
Retested on this PR's Windows artifact (`elmerfem-nompi-921_merge-26.2-3b8e1a8c2-20260912_Windows-AMD64.zip`) with the same 26-case wrapper matrix as the #912 report (Windows 11, non-build machine, subprocess-driven, `USE DefUtils` UDF, DLL on disk as ground truth).

**The three rows this PR is for all flipped green.** Nothing else in the matrix moved.

| case | `PATH` | cwd | env | devel `8ae4a05b8` (#912) | PR #921 `3b8e1a8c2` |
|---|---|---|---|---|---|
| P1 | `<EH>\bin` **not** on PATH | the UDF's directory | none | exit 1, no DLL, no message | **exit 0, `UDF.dll`** |
| P2 | not on PATH | `<EH>\bin` | none | exit 1 — `collect2: cannot find 'ld'` | **exit 0, `UDF.dll`** |
| P3 | not on PATH | UDF dir | `ELMER_HOME=<EH>` | exit 1, silent | **exit 0, `UDF.dll`** |

The other 23 rows are byte-for-byte the #912 results: no-env absolute/relative/`.\`/via-PATH invocation, junction, spaces, Latin-1 path, `ELMER_LIB`-alone — all exit 0; the designed failures (`ELMER_LIB` at the runtime dir, broken source, stale `ELMER_HOME`/`ELMER_LIB`, wrapper copied out of its tree without env) still exit 1 for the same reasons; characters outside the ANSI code page still 127. So with #912 + #921 the compiler wrappers now literally need no environment: unpack the zip anywhere, run `elmerf90` from anywhere.

One corner from reading the change, then measured: when `PATH` is **entirely absent** from the process environment, `GetEnvironmentVariableA("PATH", NULL, 0)` returns 0 and the prepend is skipped, so that run is back to the silent exit 1 (a `PATH` of just `C:\Windows\System32` is enough to make it work). Rare — a constructed environment, not a shell — but `SetEnvironmentVariableA("PATH", exe_path)` in the `buffer_size == 0` branch would close it for one line.

This also retires the "failure got quieter" point from #858 — the failing grandchild no longer exists. Nice.

mmuetzel:
Thanks for testing!

You're right. I'm setting a new value if `PATH` was empty or absent also in #919. So, why not do it here, too?
It's likely not going to work in either case because (almost) every Windows executable depends on `kernel32.dll` in one form or the other. And not having `C:\Windows\System32` (or wherever Windows is installed) somewhere on the `PATH` will probably cause all sorts of problems. (But maybe, that will change with some version of the OS. However unlikely that is.)

I've also modified it to get rid of the `memmove` by writing at the right position in the buffer in the first place.
And I added a check whether the OS ran out of memory for allocating that buffer.

king-aj3:
Retested the revised artifact (`elmerfem-nompi-921_merge-26.2-4a4b870e0-20260913_Windows-AMD64.zip`). The `PATH`-absent hardening works — and it works fully, not "attempt anyway":

| environment (no `ELMER_*`, cwd = the UDF's directory) | previous head `3b8e1a8c2` | revised `a862f65ad` |
|---|---|---|
| `PATH` absent entirely | exit 1, no DLL, silent | **exit 0, `UDF.dll`** |
| `PATH` present but empty | exit 0 (the size query returns 1 for the terminator, so the prepend already ran) | exit 0 |
| `PATH` = `C:\Windows\System32` only | exit 0 | exit 0 |

The reason it does not fail on `kernel32.dll`: that is a Known DLL, loaded from `System32` regardless of `PATH` — so with `PATH` set to just the wrapper's directory, gfortran, `f951` and `collect2` all find everything they need. The full 26-case wrapper matrix on this head is byte-identical to the previous one (the three bin-off-PATH rows green, the designed failures unchanged). Nothing left from this side.

mmuetzel:
Thanks for testing again. 👍

I wasn't aware that there was a thing as "Known DLL". But thinking about it, it would be a huge security hole if everyone would be able to shadow that `kernel32.dll` from MS by just putting a library with that name somewhere early in the `PATH`.

I amended the PR with an adapted comment. No functional changes (apart from rebasing on a current head).

mmuetzel:
@raback: This should be ready for merging.

raback:
Thanx Markus and King again! It is nice to follow such a high level of coding and testing.

PR #926, opened 2026-09-14
Feat hypre gpu support

dancaseb:
Some minor fixes for Hypre:
1. It is possible to compile hypre without the --enable-unified-memory flag, replacing it with --with-cuda flag. If --with-cuda flag is used, Elmer now correctly binds each mpi process to it's own GPU. Before all MPI processes would be binded to GPU 0, causing the program to hang.
2. Multi GPU runs are now possible thanks to GPU binding.
3. In the AMS setup, HYPRE_AMSSetAlphaAMGOptions and HYPRE_AMSSetBetaAMGOptions use different interpolation type (6). Interpolation type 0 works only on CPU and was causing crashes before. This could possibly be parametrized if there is a use case for it.


raback:
Thanx Daniel! Nice work.

